PR #69, opened 2021-08-06
Add IgnoreList

chevdor:
I saw this repo uses `main` vs `master` so this ~~blacklist~~ is called `ignoreList` (mostly but not exclusively because people searching for the feature will likely search for the ~~blacklist~~ term).

The ~~blacklist~~ ignore list can be populated via `FAUCET_IGNORE_LIST` as documented in the readme.

For instance (those accounts are mine and can be added to the ignore list...)
```
FAUCET_IGNORE_LIST="@alice:matrix.org","@bob:domain.com"
```

NOTE: The sum of all the ENV should not exceed `32767`. That's around 600 ignored addresses if we keep some room for a few other ENV. We can think about handling a longer list if required once we get that far.


src/bot/ignorellist.ts
```diff
@@ -0,0 +1,4 @@
+export function isIgnored (account: string): boolean {
+  const ignoreList: string[] = (process.env.FAUCET_IGNORE_LIST ?? '').replace(/\s/g, '').split(',') ?? [];
+  return ignoreList.includes(account);
```
wpank, line 3: One should also potentially convert the inputted address for the specific ss58 encoding of the chain the faucet is run for. 

https://polkadot.js.org/docs/keyring/start/ss58#how-is-the-encoding-done

I think someone can just input a different ss58 representation of the address to get around the ignore list.

wpank: I think this can be done by:

```javascript
import { decodeAddress } from '@polkadot/keyring';

export function isIgnored (account: string): boolean {
    try {
      decodeAddress(arg0);
    } catch (e) {
      logger.warn(`${account} is not in the proper format.`);
      return false;
    }
  const ignoreList: string[] = (process.env.FAUCET_IGNORE_LIST ?? '').replace(/\s/g, '').split(',') ?? [];

  return ignoreList.includes(decodeAddress(account));
}
```

wpank: This of course will then assume that the ignore list is ss58 specific for the given chain.

chevdor: Yes, this PR aism at a simple solution to have at least something in place.
Since we rely on matrix accounts and those are (slightly) harder to produce than a new address, a filter on the address could be done however a matrix user can also simply be banned from the faucet channel.

lovelaced (approved):

lovelaced (changes requested):
This variable will also need to be added to the k8s chart.

chevdor:
So I think I added the new variable where it should but I was not able to test locally in my cluster:
```
helm install -n testing --dry-run --debug --generate-name ./kubernetes/faucetbot/
install.go:173: [debug] Original chart version: ""
install.go:190: [debug] CHART PATH: /Users/Shared/projects/substrate-matrix-faucet/kubernetes/faucetbot

Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
helm.go:88: [debug] unable to recognize "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
unable to build kubernetes objects from release manifest
```

src/bot/ignorellist.ts
chevdor: I added your suggestion. Totally untested yet.

chevdor: I switched to ignoring Matrix accounts, it ended up easier and this is stronger.

chevdor:
I did test the following:
- [x] local run (server + bot)
- [x] helm template generation
- [x] helm template install (I skipped the ServiceMonitor though)

I need to fix a last issue. Do not merge.

src/bot/ignorellist.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Check whether an account is in the ignore list
+ * @param matrixAccount The matrix account
+ * @returns true/false
+ */
+export function isIgnored (matrixAccount: string): boolean {
+  const ignoreList: string[] = process.env.FAUCET_IGNORE_LIST?.split(',') ?? [];
```
AndreiEres, line 7: I see we use `getEnvVariable` to get vars

AndreiEres (commented):
Hello!
I reviewed PR and added a few comments. Could you have a look?

README.md
```diff
@@ -48,6 +48,13 @@ NETWORK_UNIT="CAN"
 ```
 ---
 
+## Ignore List / Blacklist
+
+A list of Matrix accounts that will be silently (but logged) ignored:
+```
+FAUCET_IGNORE_LIST="@alice:matrix.org","@bob:domain.com"
```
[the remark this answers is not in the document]
chevdor: This is fine, you can write it both ways. I just found the example easier written this way. You can provide is as `FAUCET_IGNORE_LIST=@alice:matrix.org,@bob:domain.com` as well.

deployment.template.yaml
```diff
@@ -24,6 +24,13 @@ spec:
             - name: BACKEND_URL
               value: "http://localhost:5555"
           imagePullPolicy: Always
+          resources:
```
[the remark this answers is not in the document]
chevdor: That can be removed if you see an issue with that. In my tools, specs without resources defined raise a warning.

src/bot/ignorellist.ts
```diff
@@ -0,0 +1,9 @@
+/**
```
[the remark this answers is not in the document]
chevdor: 100% ! Fixed.

deployment.template.yaml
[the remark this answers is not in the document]
AndreiEres: Can we add it via another PR and rebase?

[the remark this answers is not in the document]
chevdor: Sure. I remove it.

README.md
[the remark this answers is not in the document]
AndreiEres: I think `FAUCET_IGNORE_LIST=@alice:matrix.org,@bob:domain.com` or `FAUCET_IGNORE_LIST="@alice:matrix.org,@bob:domain.com"` looks simply and without brainstorming how it works. Could you fix it?

AndreiEres:
@chevdor Do you need any help to complete this PR? I can just fix remaining remarks and merge PR if you have no time.

chevdor:
@AndreiEres I am on other topics atm so please DO feel free to apply the fixes you want. My priority is only to have this feature available **in some form** so we have a simple solution that can be easily put in place quickly against abuses. We can refactor the PR or you may consider merging it as is and provide an improvement PR on top.

AndreiEres (approved):

src/bot/ignoreList.ts
```diff
@@ -1,9 +1,14 @@
...
+
 /**
  * Check whether an account is in the ignore list
- * @param matrixAccount The matrix account
- * @returns boolean
  */
-export function isIgnored (matrixAccount: string): boolean {
-  const ignoreList: string[] = process.env.FAUCET_IGNORE_LIST?.split(',') ?? [];
-  return ignoreList.includes(matrixAccount);
+export function isIgnored ({
```
chevdor, line 9: Let's consider we have a big list of ignored accounts. Here you pass the huge list over and over again (for every faucet request). That does not sound optimum.

chevdor (commented):
I cannot approve my own PR but the changes LGTM beside the point I commented.

src/bot/ignoreList.ts
AndreiEres: You are right. So better to check it without extra functions.

src/bot/index.ts
```diff
@@ -6,12 +6,14 @@ import * as mSDK from 'matrix-js-sdk';
 import { isDripSuccessResponse } from '../guards';
 import type { BalanceResponse, DripResponse, EnvNameBot, EnvVar } from '../types';
 import { checkEnvVariables, getEnvVariable, logger } from '../utils';
+import { isIgnored } from './ignoreList';
```
chevdor, line 9: I think this is no longer needed now

chevdor (commented):
Looking good now, besides the code that is no longer needed.

lovelaced (approved):

AndreiEres:
@chevdor Let's merge?

chevdor:
I am fine merging it. I did not test this new version though. 
@AndreiEres did you test it ?

AndreiEres:
> @AndreiEres did you test it ?

No, And I don't know how.

chevdor:
DO NOT MERGE (tests in progress...)

## Fails
I am testing and the recent changes make it break in some scenarii such as:
- [ ] `!drip` breaks [here](https://github.com/paritytech/substrate-matrix-faucet/blob/wk-blacklist/src/bot/index.ts#L106) when `arg0` is undefined

## Pass
- ✅ normal drip (`!drip 5CwPxumBRDLkP7VQEYzhwoYw6AP4FNmRM7G1pj7Atj6dEzgY`)
- ✅ drip with amount (`!drip 5CwPxumBRDLkP7VQEYzhwoYw6AP4FNmRM7G1pj7Atj6dEzgY 2`)
- ✅ invalid command `!dripppy`
- ✅ `!help`
- :x: see #80

PR #74, opened 2021-09-06
Provide better error messaging on dripping

AndreiEres:


src/types.ts
```diff
@@ -34,8 +34,8 @@ export interface BalanceResponse {
 }
 
 export interface DripResponse {
```
joao-paulo-parity, line 36: it makes more sense to model this as a union type

`export type DripResponse = Error | { hash: string }`

That way you would be able to handle it more safely in other places

src/server/actions.ts
```diff
@@ -53,31 +54,33 @@ export default class Actions {
...
 
-      logger.info('💸 sending tokens');
+    try {
+      if (!this.account) throw new Error('account not ready');
 
       const dripAmount = Number(amount) * 10 ** decimals;
+      const api = await this.getApiInstance();
       const transfer = api.tx.balances.transfer(address, dripAmount);
 
+      logger.info('💸 sending tokens');
```
joao-paulo-parity, line 68: it makes more sense to log this **before** calling api.tx.balances.transfer

joao-paulo-parity (approved):
Some _optional_ code improvements suggested. Feel free to merge.

src/server/actions.ts
AndreiEres: We call it when we sign and send.

src/types.ts
AndreiEres: What do you mean?

src/server/actions.ts
joao-paulo-parity: It would be a good idea to divide this log in two messages. If api.tx.balances.transfer throws an exception right now, you never get to this line so it's less clear what happened from the logs.

src/types.ts
joao-paulo-parity: Dealing with this DripResponse type is not good. Example from the code:

```typescript
      const message = res.data.hash
        ? `Sent ${sender} ${dripAmount} ${unit}s. Extrinsic hash: ${res.data.hash}`
        : (res.data.error || 'An unexpected error occured, please check the server logs');
```

Since it has two optional fields, it might lead to bugs where neither is defined. Also handling it is inconsistent: here you are **choosing** for `data.hash` and, if it does not exist, checking for `data.error`; however, there's no reason why it could not be `data.error` first and `data.hash` after.

When a DripResponse has `{ hash: "something", error: "something" }`, what does it mean? Was it an error or success? Really, it should never be in two states ("error" and "success") at the same time, but it's currently allowed because the error and success information are encoded in the same type.

If it's defined as distinct types, handling it would be safer and saner:

```typescript
      const message = res instanceof Error
        ? res.message
        :  `Sent ${sender} ${dripAmount} ${unit}s. Extrinsic hash: ${res.data.hash}`
```

If you want to apply this change, I suggest replacing the type with the signature I provided and start fixing the errors which appear.

AndreiEres: Fair enough. Firstly it was separated like `{hash} | {error}`.
But I can't user Error directly, because it throws.

PR #79, opened 2021-09-22
Upgrade polkadot dependencies

PierreBesson:
To fix Rococo faucet.


PierreBesson:
Don't merge yet. The build doesn't pass due an error.

PierreBesson:
I'm not sure if my solution to the typescript error is the most elegant but it should work.

gabreal (approved):

Issue #82, opened 2021-10-01
Add support for websocket and subscriptions


chevdor:
Currently, for some reasons, the server only support `http`. That prevents from benefiting from subscriptions.

ref. #67

PierreBesson:
I think migrating to the Websocket provider might make the Faucet more resistant to the disconnects that happen when we upgrade RPC nodes which are currently breaking the Westend and Rococo faucets frequently. So it would be nice to see this addressed.

PierreBesson:
In my opinion this feature is very important. WS connections have been the standard for connecting to RPC nodes for many years by now.

mordamax:
@PierreBesson thanks for emphasising it! That seems to be related/duplicate of this issue https://github.com/paritytech/substrate-matrix-faucet/issues/84 right?
if so, @parutger might be able to provide some timeline on fixing this

PierreBesson:
Those are separate issues but it's possible that part of the answer to fix #84 is to switch from HTTP to WS.

mutantcornholio:
As of today, all faucets are using websocket RPCs

Issue #80, opened 2021-10-01
Low balance is not checked/detected


chevdor:
(ignore the bot name below.. I am recycling :))

```
polkabot-dev
The faucet has 0.68149997781 WNDs remaining.
Will / Chevdor
!drip 5CwPxumBRDLkP7VQEYzhwoYw6AP4FNmRM7G1pj7Atj6dEzgY
polkabot-dev
Sent @chevdor:matrix.parity.io 1 WNDs. Extrinsic hash: 0xaf1888d5812b8dc0b37bc8bc57ea594712e5ec70761fe31c8ae19ec1a28af5b8
```

results in a [failed extrinsic](https://westend.subscan.io/extrinsic/0xaf1888d5812b8dc0b37bc8bc57ea594712e5ec70761fe31c8ae19ec1a28af5b8) obviously, note this is caught no where and the bot proudly shows in the chat that 1 WND was sent !

chevdor:
Can also be tested with a number that is bigger than the current balance:
```
!drip 5CwPxumBRDLkP7VQEYzhwoYw6AP4FNmRM7G1pj7Atj6dEzgY 99999999
```

PR #83, opened 2021-10-01
Add prometheus metrics (including faucet balance)

### LINKED ISSUE #67: Feature request: metric for balance amount
It would be cool and useful to have a prometheus metric for the amount of tokens left in the faucet. This way we could create an alert when the faucet needs refilling, and can also see how many "people" are requesting tokens.

chevdor:
This PR adds new Prometheus metrics:
- balance
- total_requests
- successful_requests

It also does:
- remove all the console.log and switch to using the logger
- log the version at start
- bump the server version
- ⚠️ move previous metrics to the new `faucet_` namespace (see example below) 

The goal was not to rewrite too much code so the following is left for other PRs:
- #80 
- #82 

```
# HELP faucet_errors_total The total amount of errors logged on the faucet backend
# TYPE faucet_errors_total counter
faucet_errors_total 0
# HELP faucet_errors_rpc_timeout The total amount of timeout errors between the faucet backend and the rpc node
# TYPE faucet_errors_rpc_timeout counter
faucet_errors_rpc_timeout 0
# HELP faucet_balance Current balance of the faucet
# TYPE faucet_balance gauge
faucet_balance 8.36929
# HELP faucet_total_requests Total number of requests to the faucet
# TYPE faucet_total_requests gauge
faucet_total_requests 0
# HELP faucet_successful_requests The total number of successful requests to the faucet
# TYPE faucet_successful_requests gauzge
faucet_successful_requests 0
```

fix #67

### DIFF
--- a/src/server/actions.ts
+++ b/src/server/actions.ts
@@ -54,8 +54,7 @@ export default class Actions {
   }
 
   /**
-   * Checks the current balance and updates the `faucetBalance` property.
-   * @returns
+   * This function checks the current balance and updates the `faucetBalance` property.
    */
   private async updateFaucetBalance () {
     if (!this.account) return;
--- a/src/server/actions.ts
+++ b/src/server/actions.ts
@@ -42,7 +42,11 @@ export default class Actions {
       this.account = keyring.addFromMnemonic(mnemonic);
 
       // TODO: Adding a subscription would be better but the server supports on http for now
-      setInterval(() => { this.updateFaucetBalance() }, balancePollIntervalMs);
+      setInterval(() => {
+        // We do want the following to just start and run
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        this.updateFaucetBalance();
+      }, balancePollIntervalMs);
     }).catch((e) => {
       logger.error(e);
       errorCounter.plusOne('other');
--- a/src/server/actions.ts
+++ b/src/server/actions.ts
@@ -45,7 +45,7 @@ export default class Actions {
       setInterval(() => {
         // We do want the following to just start and run
         // eslint-disable-next-line @typescript-eslint/no-floating-promises
-        this.updateFaucetBalance();
+        this.updateFaucetBalance().catch(console.error);
       }, balancePollIntervalMs);
     }).catch((e) => {
       logger.error(e);
--- a/src/server/actions.ts
+++ b/src/server/actions.ts
@@ -2,6 +2,7 @@ import { ApiPromise } from '@polkadot/api';
 import Keyring from '@polkadot/keyring';
 import { KeyringPair } from '@polkadot/keyring/types';
 import { HttpProvider } from '@polkadot/rpc-provider';
+import BN from 'bn.js';
 import dotenv from 'dotenv';
 import { DripResponse } from 'src/types';
 
@@ -28,6 +29,7 @@ const rpcTimeout = (service: string) => {
 export default class Actions {
   api: ApiPromise | undefined;
   account: KeyringPair | undefined;
+  #faucetBalance: number | undefined;
 
   constructor () {
     this.getApiInstance().then(() => {
@@ -37,10 +39,25 @@ export default class Actions {
       // if we don't wait we'll get an error "@polkadot/wasm-crypto has not been initialized"
       const keyring = new Keyring({ type: 'sr25519' });
       this.account = keyring.addFromMnemonic(mnemonic);
+
+      // Adding a subscription would be better but the server supports on http for now
+      const _faucetBalancePollingId = setInterval(this.updateFaucetBalance.bind(this), 6000);
     }).catch((e) => {
       logger.error(e);
       errorCounter.plusOne('other');
     });
+  }
+
+  /**
+   * This function checks the current balance and updates the `faucetBalance` property.
+   * @returns
+   */
+  private async updateFaucetBalance() {
+    const api = await this.getApiInstance();
+    if (!this.account) return;
+    const { data: balances } = await api.query.system.account(this.account.address);
+    const precision = 5;
+    this.#faucetBalance = balances.free.toBn().div( new BN( 10 ** (decimals - precision))).toNumber()/ 10 ** precision;
   }
 
   async getApiInstance (): Promise<ApiPromise> {
@@ -52,6 +69,10 @@ export default class Actions {
 
     await this.api.isReady;
     return this.api;
+  }
+
+  public getFaucetBalance() {
+    return this.#faucetBalance;
   }
 
   async sendTokens (address: string, amount: string): Promise<DripResponse> {
--- a/src/server/actions.ts
+++ b/src/server/actions.ts
@@ -41,9 +41,8 @@ export default class Actions {
       const keyring = new Keyring({ type: 'sr25519' });
       this.account = keyring.addFromMnemonic(mnemonic);
 
-      // Adding a subscription would be better but the server supports on http for now
-      this.updateFaucetBalance();
-      const _faucetBalancePollingId = setInterval(this.updateFaucetBalance.bind(this), balancePollIntervalMs);
+      // TODO: Adding a subscription would be better but the server supports on http for now
+      setInterval(() => this.updateFaucetBalance.bind(this), balancePollIntervalMs);
     }).catch((e) => {
       logger.error(e);
       errorCounter.plusOne('other');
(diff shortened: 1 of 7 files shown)


src/server/actions.ts
```diff
@@ -37,12 +40,31 @@ export default class Actions {
...
+      }, balancePollIntervalMs);
     }).catch((e) => {
       logger.error(e);
       errorCounter.plusOne('other');
     });
   }
 
+  /**
+   * This function checks the current balance and updates the `faucetBalance` property.
+   * @returns
```
AndreiEres, line 58: ✂️

src/server/actions.ts
```diff
@@ -37,12 +40,31 @@ export default class Actions {
...
+        this.updateFaucetBalance();
+      }, balancePollIntervalMs);
     }).catch((e) => {
       logger.error(e);
       errorCounter.plusOne('other');
     });
   }
 
+  /**
+   * This function checks the current balance and updates the `faucetBalance` property.
```
AndreiEres, line 57:
```suggestion
   * Checks the current balance and updates the `faucetBalance` property.
```

src/server/actions.ts
```diff
@@ -28,6 +30,7 @@ const rpcTimeout = (service: string) => {
 export default class Actions {
   api: ApiPromise | undefined;
   account: KeyringPair | undefined;
+  #faucetBalance: number | undefined;
```
AndreiEres, line 33: What the purpose of using `#` here?

And can we declare it as optional with `?` instead of implicitly show undefined?

src/server/actions.ts
```diff
@@ -37,12 +40,31 @@ export default class Actions {
...
     });
   }
 
+  /**
+   * This function checks the current balance and updates the `faucetBalance` property.
+   * @returns
+   */
+  private async updateFaucetBalance () {
+    const api = await this.getApiInstance();
+    if (!this.account) return;
```
AndreiEres, line 62:
```suggestion
    if (!this.account) return;
    
    const api = await this.getApiInstance();
```

src/server/actions.ts
```diff
@@ -37,12 +40,31 @@ export default class Actions {
...
+    const { data: balances } = await api.query.system.account(this.account.address);
+    const precision = 5;
```
AndreiEres, line 64: Better move it to constant out of scope

src/server/actions.ts
```diff
@@ -2,6 +2,7 @@ import { ApiPromise } from '@polkadot/api';
 import Keyring from '@polkadot/keyring';
 import { KeyringPair } from '@polkadot/keyring/types';
 import { HttpProvider } from '@polkadot/rpc-provider';
+import BN from 'bn.js';
```
AndreiEres, line 5: Why we need it?

src/types.ts
```diff
@@ -23,6 +23,13 @@ export type EnvNameBot = 'BACKEND_URL'|
  'FAUCET_IGNORE_LIST'
 ;
 
+export interface MetricsDefinition {
+    meta: {
+      prefix: string,
+    },
+    data: { [id: string] : number }
```
AndreiEres, line 30:
```suggestion
    data: Record<string, number>
```

src/server/actions.ts
```diff
@@ -37,12 +40,31 @@ export default class Actions {
       // if we don't wait we'll get an error "@polkadot/wasm-crypto has not been initialized"
       const keyring = new Keyring({ type: 'sr25519' });
       this.account = keyring.addFromMnemonic(mnemonic);
+
+      // TODO: Adding a subscription would be better but the server supports on http for now
+      setInterval(() => {
```
AndreiEres, line 45: I don't like the idea to do it in initialiser. Maybe we can do it better?

src/server/index.ts
```diff
@@ -12,33 +13,68 @@ import Storage from './storage';
...
+ * Simplistic function to generate a prometheus metrics.
+ * TODO: Switch to prom-client
+ * @param name Name of the metric
+ * @param type Type
+ * @param value Value
+ * @param help Help test
+ * @param voidIfUndefined Whether we render it even if null/undefined
+ * @returns string
+ */
+function getMetrics (name: string, type: string, value: number | string | undefined, help = '', voidIfUndefined = false): string {
```
AndreiEres, line 41: Better pass object with that params

src/server/index.ts
```diff
@@ -12,33 +13,68 @@ import Storage from './storage';
...
+
+  return result;
+}
 
 const app = express();
 app.use(bodyParser.json());
 
 checkEnvVariables(envVars);
 
-const port = getEnvVariable('PORT', envVars) as string;
```
AndreiEres, line 21: But it's a string, isn't it?

src/server/index.ts
```diff
@@ -12,33 +13,68 @@ import Storage from './storage';
...
 app.get('/metrics', (_, res) => {
-  res.end(`# HELP errors_total The total amount of errors logged on the faucet backend
-# TYPE errors_total counter
-errors_total ${errorCounter.total()}
-# HELP errors_rpc_timeout The total amount of timeout errors between the faucet backend and the rpc node
-# TYPE errors_rpc_timeout counter
-errors_rpc_timeout ${errorCounter.getValue('rpcTimeout')}
-`
-  );
+  const errors_total = getMetrics('errors_total', 'counter', errorCounter.total(), 'The total amount of errors logged on the faucet backend');
```
AndreiEres, line 66: Why `snake_case`?

src/server/index.ts
```diff
@@ -12,33 +13,68 @@ import Storage from './storage';
...
+  const errors_rpc_timeout = getMetrics('errors_rpc_timeout', 'counter', errorCounter.getValue('rpcTimeout'), 'The total amount of timeout errors between the faucet backend and the rpc node');
+
+  const balance = getMetrics('balance', 'gauge', actions.getFaucetBalance(), 'Current balance of the faucet', true);
+
+  const total_requests = getMetrics('total_requests', 'gauge', metrics.data.total_requests, 'Total number of requests to the faucet');
+  const successful_requests = getMetrics('successful_requests', 'gauzge', metrics.data.success_requests, 'The total number of successful requests to the faucet');
+
+  res.end(`${errors_total}${errors_rpc_timeout}${balance}${total_requests}${successful_requests}`);
```
AndreiEres, line 74: I think joined array looks more reasonable here
```
[
  getMetrics('errors_total', 'counter', errorCounter.total(), 'The total amount of errors logged on the faucet backend'),
  ...
].join('')

File: src/server/index.ts [hunk 13]
```diff
@@ -12,33 +13,68 @@ import Storage from './storage';
...
 });
 
 const createAndApplyActions = (): void => {
-  const actions = new Actions();
```
AndreiEres, line 40: But here we create and apply actions

src/server/index.ts
```diff
@@ -53,16 +89,20 @@ const createAndApplyActions = (): void => {
 
   app.post<unknown, DripResponse, BotRequestType>('/bot-endpoint', (req, res) => {
     const { address, amount, sender } = req.body;
+    metrics.data.total_requests++;
 
     storage.isValid(sender, address).then(async (isAllowed) => {
+      const privileged = sender.endsWith(':matrix.parity.io') || sender.endsWith(':web3.foundation');
```
AndreiEres, line 95: Looks like a change from another PR. Hasn't rebased?

src/server/actions.ts
AndreiEres: But wait. How it works? How actions can have a balance? Why don't we use own class for balance or something like that?

src/server/index.ts
```diff
@@ -12,33 +13,68 @@ import Storage from './storage';
 
 dotenv.config();
 const storage = new Storage();
+const actions = new Actions();
+
+const metrics: MetricsDefinition = {
```
AndreiEres, line 18: Maybe we can define it in another module?

src/server/index.ts
```diff
@@ -53,16 +89,20 @@ const createAndApplyActions = (): void => {
 
   app.post<unknown, DripResponse, BotRequestType>('/bot-endpoint', (req, res) => {
     const { address, amount, sender } = req.body;
+    metrics.data.total_requests++;
```
AndreiEres, line 92: Maybe ok, but looks a bit imperatively

src/server/actions.ts
chevdor: `#` makes the field private: [reference](https://www.typescriptlang.org/docs/handbook/classes.html#ecmascript-private-fields)

chevdor: > And can we declare it as optional with ? instead of implicitly show undefined?

We can but I kept a style similar to the existing code.

src/server/index.ts
chevdor: The question should be whether it works or not. We can delay this PR and make it perfect but it leaves us without metrics in the meantime. Or we start with a working version and you can refactor it if you wish ?
I do agree with you though. The change is simple, it just needs to be tested again then.

src/server/index.ts
chevdor: I just did.

src/server/index.ts
chevdor: The general architecture does not make it super convenient to report metrics. Here we get a metric that could help and it does not cost much.

chevdor:
I agree with most comments but those require another round of testing for which I don't have much time now.
So I propose to merge this PR unless we find critical issues, that will allow having some important metrics already.
We can then follow up with https://github.com/paritytech/substrate-matrix-faucet/issues/85

AndreiEres:
> I agree with most comments but those require another round of testing for which I don't have much time now. So I propose to merge this PR unless we find critical issues, that will allow having some important metrics already. We can then follow up with #85

ok

AndreiEres (approved):

src/server/actions.ts
chevdor: fixed

PR #89, opened 2021-10-06
Update linters

AndreiEres:


chevdor (changes requested):
Very Nice! Looking forward to have this in!

I feel we are mixing up **linting** vs **formatting** though. While linting may (and does here) include formatting, it should be kept separate, especially because it allows to fail faster. This is also represented by the tools: you lint with eslint and format with prettify.

I would suggest to keep those steps separated and run the formatting/prettifying first.

The git hooks will make it super convenient but we cannot rely on them (users may decide against installing them). So I think we could make CI fail faster if we first check the formatting and then run the linter.

So I would suggest to do the following in package.json:
- (removing `lint:fix` is ok, since `yarn lint --fix` does the same)
- rename what you called `fix` to `format` or `prettify`. This one should only check the formatting and change nothing
- add a `format:fix` (your current `fix`) that actually changes the formatting
- add a `check` script that changes nothing but run the formatting + linting checks

AndreiEres:
> * (removing `lint:fix` is ok, since `yarn lint --fix` does the same)

Unfortunately we have eslint rules like for imports sorting, that can be fixed only via eslint. 

So we can separate it doing lint:fix, format:fix.

chevdor:
I think it is important for the dev to have the choice to check only, and/or fix as well.

AndreiEres:
> Yep, that sounds perfect to me.

Updated

Issue #93, opened 2021-10-07
Add/check readyness probe


chevdor:
There are issues everytime we redeploy:

![image](https://user-images.githubusercontent.com/738724/136387084-8c608c53-904f-4146-a461-53f7cb9c6e04.png)

I think that can be solved easily with a readyness probe and maybe a delay.
I issue is very likely that we THINK the service is fully up whereas it is still starting and not fully ready yet.

So in short, k8s should not switch to the new pods as quick as it does and let the new ones get ready

chevdor:
The issue may well be in [this part](https://github.com/paritytech/substrate-matrix-faucet/blob/main/src/server/index.ts#L71-L73) of the code actually.

PR #94, opened 2021-10-08
Add readiness probe

AndreiEres:
Fix https://github.com/paritytech/substrate-matrix-faucet/issues/93

### DIFF
--- a/deployment.template.yaml
+++ b/deployment.template.yaml
@@ -22,7 +22,7 @@ spec:
                   name: $CI_PROJECT_NAME-config
                   key: matrix-access-token
             - name: BACKEND_URL
-              value: "http://localhost:5555"
+              value: 'http://localhost:5555'
           imagePullPolicy: Always
         - name: $CI_PROJECT_NAME-server
           image: paritytech/$DOCKER_IMAGE-server:$DOCKER_TAG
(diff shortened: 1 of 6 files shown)


deployment.template.yaml
```diff
@@ -32,11 +32,18 @@ spec:
                 secretKeyRef:
                   name: $CI_PROJECT_NAME-config
                   key: mnemonic
+          readinessProbe:
+            httpGet:
+              path: /ready
+              port: 5555
+            failureThreshold: 1
+            periodSeconds: 30
+            initialDelaySeconds: 60
```
chevdor, line 41: I think we can keep it tight here.

```suggestion
            periodSeconds: 10
            initialDelaySeconds: 60
```

kubernetes/faucetbot/templates/deployment.yaml
```diff
@@ -65,13 +65,20 @@ spec:
                   key: {{ $key }}
                   name: {{ $.Values.chain.name }}-server-config
           {{- end }}
+          readinessProbe:
+            httpGet:
+              path: /ready
+              port: 5555
+            failureThreshold: 0
+            periodSeconds: 30
+            initialDelaySeconds: 60
```
chevdor, line 74: same as above

src/server/index.ts
```diff
@@ -68,6 +68,10 @@ checkEnvVariables(envVars);
 
 const port = getEnvVariable('PORT', envVars) as number;
 
+app.get('/ready', (_, res) => {
+  res.send('Faucet backend is ready.');
+});
+
```
chevdor, line 74: Doing this does not give us much more than the previous `/health` endpoint.
Here it would be great to have a success (200) response ONLY if we can query the API (any call).

chevdor (changes requested):
I think we need at least some kinda real check for `/ready`.
We can do it cheap and slow with performing an API call in the endpoint.

Or better logging the time of the last success call to the API and considering we are ready if the time since this last good call is under a threshold.

chevdor:
@azerella CI seems to be failing since your `Unused import` change.

azerella:
> @azerella CI seems to be failing since your `Unused import` change.

Looks like the runners were just borked, re-running them seems to have fixed it

chevdor (approved):

PR #99, opened 2021-10-11
Refactoring express route project structure

### LINKED ISSUE #98: Refactor routes and organise server project structure
We load _all_ of the routes into `server/index.ts` currently and it's getting quite cluttered, we can split a lot of this out into separate routes

azerella:
Fixes #98

### DIFF
--- a/src/server/constants/prometheus.ts
+++ b/src/server/constants/prometheus.ts
@@ -1,0 +1,14 @@
+import { MetricsDefinition } from '../../types';
+
+export const metrics: MetricsDefinition = {
+  data: {
+    balance: 0,
+    errors_rpc_timeout: 0,
+    errors_total: 0,
+    success_requests: 0,
+    total_requests: 0,
+  },
+  meta: {
+    prefix: 'faucet',
+  },
+};
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -3,175 +3,21 @@ import dotenv from 'dotenv';
 import express from 'express';
 
 import * as pkg from '../../package.json';
-import { isDripSuccessResponse } from '../guards';
-import type {
-  BalanceResponse,
-  BotRequestType,
-  DripResponse,
-  MetricsDefinition,
-} from '../types';
 import { checkEnvVariables, getEnvVariable, logger } from '../utils';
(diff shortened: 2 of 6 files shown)


src/server/constants/prometheus.ts
```diff
@@ -0,0 +1,14 @@
+import { MetricsDefinition } from '../../types';
+
+export const metrics: MetricsDefinition = {
+  data: {
+    balance: 0,
```
chevdor, line 5: I don't think this is a good idea to initialize the balance with 0.
It will lead to prometheus potentially reporting this and trigger alerts about the balance being 0 whereas it currently is not empty.

chevdor (approved):
Beware that @AndreiEres is bringing changes in parts you touched and I think his code should be merged first. There will likely be a few conflicts afterwards.

chevdor (changes requested):
Initializing the balance at 0 is not good IMO

src/server/constants/prometheus.ts
azerella: Ah this was simply copied from what it already is, I'm not too confident in what else this should be? Might need @AndreiEres to weigh in here

chevdor: The current code skips retuning the metrics at all until a real value is known (it does not return 0 unless it is really 0).
Not sure what Prometheus' defaults are but `null` or `undefined` would be better.

AndreiEres (approved):
LGTM, but double check before merging please

src/server/constants/prometheus.ts
chevdor: I would disagree with that, becaue if we merge this PR with the balance initializing at 0, we will get false alarms about the balance being too low. We do not have the issue today so the PR in the current state would introduce a regression. 

We don't really need another PR to fix regressions introduced in a PR if we are aware of them.

azerella: > We do not have the issue today so the PR in the current state would introduce a regression.

I don't understand that considering this is what the existing code already does? Are you saying we haven't deployed this code yet?

mordamax:
Closing in favor of #145, as this is outdated

PR #103, opened 2021-10-14
Introducing local dev hot-reloading

azerella:
This is as small a PR as I could possibly make it @AndreiEres - if we merge this then I'll do follow-up PRs including updating the `README.md`


Dockerfile-bot
```diff
@@ -1,6 +1,23 @@
...
 WORKDIR /bot
+
 COPY package.json .
-RUN yarn install
-COPY . .
-CMD yarn bot
+COPY yarn.lock .
+COPY babel.config.json .
+COPY tsconfig.json .
+COPY src/ ./src
```
chevdor, line 11: Isn't it easier and less error prone to copy all and filter out in the `.dockerignore` file ?

Dockerfile-server
```diff
@@ -1,6 +1,23 @@
...
 WORKDIR /backend
+
 COPY package.json .
-RUN yarn install
-COPY . .
-CMD yarn backend
+COPY yarn.lock .
+COPY babel.config.json .
+COPY tsconfig.json .
+COPY src/ ./src
```
chevdor, line 11: Isn't it easier and less error prone to copy all and filter out in the `.dockerignore` file ?

.dockerignore
```diff
@@ -1,2 +1,3 @@
 build
 node_modules
+__tests__
```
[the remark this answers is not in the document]
chevdor: Why the `__` ? Once we have tests they can go to the `test` folder.

chevdor (changes requested):

.dockerignore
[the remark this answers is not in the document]
azerella: No real reason, just a naming convention I like to put tests in, makes it really obvious/easy to detect, we can use `test` if everyone prefers

Dockerfile-bot
azerella: I wouldn't say it's easier doing that, it's just _another_ way to do it. 

Personally, I prefer this approach as it's explicit and IMO very obvious what I'm coping over without having to cross check 2 files. I also suspect it would be quite unlikely for us to want to copy over additional files in the future that aren't in the `src` folder, this somewhat forces us to keep organised

azerella: But that's exactly what we were doing before? You were copying literally everything and relying on the '.dockerignore' file...

I don't think going back and forth on something so trivial is productive, given our time zones - I'm in Australia.

If you truely feel like copying everything like before was better than being explicit then please modify this PR, otherwise it's going to take us months to get anything done 😄

azerella: Any updates here @chevdor ? I'm hoping to get this through ASAP so that we can start improving other things and adding a test suite

azerella: @chevdor @AndreiEres

azerella:
> @azerella It does the same but in more consistent way #105

Let's add a `nodemon.json` config to be more explicit then merge away

azerella:
Closing this in favour of https://github.com/paritytech/substrate-matrix-faucet/pull/105

PR #106, opened 2021-11-01
Validate and parse environment variables

### LINKED ISSUE #87: Use confmgr
We could improve the ENV and settings management using [cnfmgr](https://gitlab.com/chevdor/confmgr).

### LINKED ISSUE #101: Add ENV for the BaseURL
The [BaseURL](https://github.com/paritytech/substrate-matrix-faucet/blob/a64611292b547f08a519fe4def40eba5ab048704/src/bot/index.ts#L67) should not be hardcoded.

azerella:
Fixes #87 
Fixes #101 

@chevdor Ended up going with `joi` here over `confmgr` as we can reuse it as a general validation library later in the future


src/utils.ts
```diff
@@ -1,79 +1,4 @@
...
-
-export function checkEnvVariables<T extends EnvNameBot | EnvNameServer>(
-  envVars: EnvVar<T>
-): void {
-  Object.entries<EnvOpt>(envVars).forEach(([env, opt]) => {
-    const value = process.env[env];
-
-    if (value === undefined) {
-      if (opt.required) {
-        console.error(`✖︎ Required environment variable ${env} not set.`);
```
azerella, line 58: `Joi` will throw an error in the console which can be used to debug what happened

chevdor:
Would you mind providing your pros/cons for `joi` over `confmgr` ?

I fail to see how `joi` brings much value and on the other hand, we'd miss a lot that it does not do compared to `confmgr`.

Leaving aside that "the author is known" :), one of the benefit of `comfmgr` is that it takes care of some check you explicitely needed to introduce in this PR.

It sounds to me that `joi` and `confmgr` are not really doing the same thing and what we really need is more about what `confmgr` is covering.

I am biaised of course and I understand that using what you know already is easier than going for what you do not know yet... 

@AndreiEres may bring a more balanced opinion on the topic. If you both want to stick with `joi`, I won't be against it but that will require more work and code to get to the same result while still missing out on neat features `confmgr` brings "for free".

azerella:
> Would you mind providing your pros/cons for `joi` over `confmgr` ?

From what I read of your API docs, `confmgr` is intended to solve for configuration or environment validation, `joi` is a _general purpose validation library_.

In the future PRs, I intend to add some middleware to _validate_ the route requests i.e querystrings, body, parameters etc.

azerella:
> > Would you mind providing your pros/cons for `joi` over `confmgr` ?
> 
> From what I read of your API docs, `confmgr` is intended to solve for configuration or environment validation, `joi` is a _general purpose validation library_.
> 
> > we'd miss a lot that it does not do compared to confmgr.
> 
> I'm not sure what we're missing that Joi doesn't handle here? I don't think we're really looking for YAML configs, regex parsing etc
> 
> If you don't like Joi we could also maybe use [zod](https://www.npmjs.com/package/zod) (zero dependencies).
> 
> In the future PRs, I intend to add some middleware to _validate_ the route requests i.e querystrings, body, parameters etc.

@chevdor Any thoughts here?

chevdor:
So it sounds like you see `joi` as benefitial since it ** is a general purpose validation library**. What else do you plan on validating ?

azerella:
> So it sounds like you see `joi` as benefitial since it ** is a general purpose validation library**. What else do you plan on validating ?

As I mentioned above

> In the future PRs, I intend to add some middleware to validate the route requests i.e querystrings, body, parameters etc.

It's good practice to validate request params, querystrings, body, headers etc - A general purpose validation library is a good tool for this. Granted we don't have many routes at the moment but they are still worth validating and obviously whatever new routes we add in the future will benefit from this

mordamax:
Closed as implemented in #133

PR #108, opened 2021-11-15
Threshold limit on the amount the faucet can dispense to an account

### LINKED ISSUE #70: Add a balance cap option
Under "some circumstances" it may be relevant to *calm down* the faucet while still allowing new comers to use it.

It would be useful to be able to optionnally setup an ENV such as `FAUCET_BALANCE_CAP` that prevents accounts with balance higher than X to successfully get drips.

azerella:
Fixes #70 

I've defaulted this to `100` but let me know if you prefer another limit

### DIFF
--- a/src/server/actions.ts
+++ b/src/server/actions.ts
@@ -84,27 +84,16 @@ export default class Actions {
     return scaledBalanceFree;
   }
 
-  async sendTokens(
-    address: string,
-    amount: string,
-    isPrivileged: boolean
-  ): Promise<DripResponse> {
+  public async isAccountOverBalanceCap(address: string): Promise<boolean> {
+    return (await this.getAccountBalance(address)) > balanceCap;
+  }
+
+  async sendTokens(address: string, amount: string): Promise<DripResponse> {
     let dripTimeout: ReturnType<typeof rpcTimeout> | null = null;
     let result: DripResponse;
 
     try {
       if (!this.account) throw new Error('account not ready');
-
-      const accountBalance = await this.getAccountBalance(address);
-      const isAccountOverBalanceCap = accountBalance > balanceCap;
-
-      if (!isPrivileged && isAccountOverBalanceCap) {
-        logger.error(
-          `⭕ Blocking faucet as this account balance is over the threshold limit of ${balanceCap}`
-        );
-
-        throw Error('Account balance is over the faucet threshold');
-      }
 
       const dripAmount = Number(amount) * 10 ** decimals;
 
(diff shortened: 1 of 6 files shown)


src/server/actions.ts
```diff
@@ -77,6 +78,19 @@ export default class Actions {
     try {
       if (!this.account) throw new Error('account not ready');
 
+      const { data } = await apiInstance.query.system.account(address);
```
AndreiEres, line 81: Maybe better to move this to a separate action and use before sending, after checking `isPrivileged` for example. Because parity and web3 members shouldn't have that restriction IMO.

AndreiEres (changes requested):
Few little changes

src/server/actions.ts
```diff
@@ -77,6 +78,19 @@ export default class Actions {
...
+      const { free: balanceFree } = data;
+      const scaledBalanceFree =
+        balanceFree.toBn().toNumber() / Math.pow(10, decimals);
```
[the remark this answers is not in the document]
azerella: IMO this is harder to read but no problem...

AndreiEres (approved):
lgtm

AndreiEres:
@chevdor what do you think, 100 is enough for the cap?

chevdor:
100 sounds like a reasonable value to get started yes.

src/server/index.ts
```diff
@@ -156,15 +161,20 @@ const createAndApplyActions = (): void => {
...
+          );
 
           // parity member have unlimited access :)
           if (!isAllowed && !isPrivileged) {
             res.send({
               error: `${sender} has reached their daily quota. Only request once per day.`,
             });
+          } else if (isAllowed && isAccountOverBalanceCap && !isPrivileged) {
+            res.send({
+              error: `${sender}'s balance is over the faucet threshold`,
```
chevdor, line 176:
```suggestion
              error: `${sender}'s balance is over the faucet's balance cap`,
```

src/server/serverEnvVars.ts
```diff
@@ -2,6 +2,12 @@ import { EnvNameServer, EnvVar } from '../types';
 
 export const envVars: EnvVar<EnvNameServer> = {
   FAUCET_ACCOUNT_MNEMONIC: { required: true, secret: true, type: 'string' },
+  FAUCET_BALANCE_CAP: {
+    default: 100,
+    required: false,
```
chevdor, line 7: What happens when the ENV does not exist ?

Scratch that, I see we have a default.

chevdor (approved):
LGTM besides the error message

PR #134, opened 2022-04-14
Added dry-run for docker build

mordamax:


alvicsam (approved):
Currently `.gitlab-ci.yml` is written not in the way we (CI/CD team) usually do. We agreed with @mordamax that I approve the PR because it works and checks necessary things and I'll fix naming and templates later

cc @paritytech/ci

PR #141, opened 2022-04-20
Fix logic of custom dripping amount (allow web3f)

### LINKED ISSUE #115: Westend faucet: web3.foundation Matrix accounts whitelisted but can't do > 1 WND drips
This is on the Matrix Westend faucet.

You can see that it worked for Chevdor (who has a parity.io account) but not for my web3.foundation account, which should also be whitelisted.

```
Bill | W3F
!drip 5EJB3zyidPPnT5bw5RgQ6EKqc6ZvrLkUjYPHStGHZArPUMUH 2000
faucet
Sent @bill:web3.foundation 1 WNDs. Extrinsic hash: 0x039e29809b1bbd7f193e707ffddc973f7e71cd9aaacbd432513a9a23e8a42a23
Bill | W3F
!balance
faucet
The faucet has 48244.067049875484 WNDs remaining.
Will / Chevdor
!help
faucet
The following commands are supported:
!balance - Get the faucet's balance.
!drip <Address> - Send WNDs to <Address>.
!help - Print this message
Will / Chevdor
!drip 5EJB3zyidPPnT5bw5RgQ6EKqc6ZvrLkUjYPHStGHZArPUMUH 2
faucet
Sent @chevdor:matrix.parity.io 2 WNDs. Extrinsic hash: 0xbbcb65410b2d20e57444f65504cfc216126fc4bebec499c4074849d137c38394
```

mordamax:
Fix #115

### DIFF
--- a/src/bot/index.ts
+++ b/src/bot/index.ts
@@ -134,7 +134,23 @@ bot.on('Room.timeline', (event: mSDK.MatrixEvent) => {
 
     // Parity users can override the drip amount by using a 3rd argument
     if (arg1 && isAccountPrivileged(sender)) {
-      dripAmount = Number(arg1) || defaultDripAmount;
+      dripAmount = Number(arg1);
+
+      // not sending these messages to matrix room, since this feature only for internal users
+      // who have access to loki logs
+      if (Number.isNaN(dripAmount)) {
+        logger.error(
+          `⭕ Failed to convert drip amount: "${arg1}" to number, defaulting to ${defaultDripAmount} ${unit}s`
+        );
+        dripAmount = defaultDripAmount;
+      }
+
+      if (dripAmount <= 0) {
+        logger.error(
+          `⭕ Drip amount can't be less than 0, got ${dripAmount}, defaulting to ${defaultDripAmount} ${unit}s`
+        );
+        dripAmount = defaultDripAmount;
+      }
     }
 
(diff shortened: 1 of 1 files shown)


src/bot/index.ts
```diff
@@ -133,8 +133,8 @@ bot.on('Room.timeline', (event: mSDK.MatrixEvent) => {
     }
 
     // Parity users can override the drip amount by using a 3rd argument
-    if (sender.endsWith(':matrix.parity.io') && arg1) {
-      dripAmount = Number(arg1);
+    if (arg1 && isAccountPrivileged(sender)) {
+      dripAmount = Number(arg1) || defaultDripAmount;
```
mordamax, line 137: converting user-input string could endup with NaN, so defaulting it to standard amount

joao-paulo-parity: I think it's better to send an error message in this case. "Defaulting to standard amount" is a server-side implementation detail which users aren't aware of.

joao-paulo-parity (changes requested):

src/bot/index.ts
mordamax: You're right, done

joao-paulo-parity (approved):
I would prefer if the dripping **didn't happen** in case of invalid input instead of defaulting to `defaultDripAmount`, but if that's your preference then I'll not object since I don't know much about what's expected of this project

PR #146, opened 2022-04-21
Make explicit deployments, rather than merging to main

### LINKED ISSUE #143: Make prod deployments to trigger only stable tags (instead of merging to main)
Make prod deployments to trigger only stable tags (instead of merging to main)

mordamax:
Fix #143

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -23,8 +23,8 @@ stages:
   - deploy
 
 .on-release-prod:                   &on-release-prod
-  only:
-    - /^v[0-9]+\.[0-9]+.*$/
+  rules:
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+.*$/'
 
 .build_lint:                        &build_lint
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -22,6 +22,10 @@ stages:
   - dockerize
   - deploy
 
+.on-release-prod:                   &on-release-prod
+  only:
+    - /^v[0-9]+\.[0-9]+.*$/
```
joao-paulo-parity, line 27: `$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+.*$/` would be more correct

mordamax: will fix 👍  for the sake of learning - could you explain why?
fyi: i copied from cla bot,

joao-paulo-parity: With the way it's implemented currently, this expression would target both branches and tags. `$CI_COMMIT_TAG` is only for tags specifically, which I infer is your intention.

PR #147, opened 2022-04-21
Mak/check low balance

### LINKED ISSUE #80: Low balance is not checked/detected

mordamax:
Fix #80

### DIFF
--- a/src/server/services/Actions.ts
+++ b/src/server/services/Actions.ts
@@ -25,18 +25,11 @@ const rpcTimeout = (service: string) => {
   }, timeout);
 };
 
-export default class Actions {
-  private static instance: Actions = new Actions();
+class Actions {
   account: KeyringPair | undefined;
   #faucetBalance: number | undefined;
 
   constructor() {
-    if (Actions.instance) {
-      throw new Error(
-        'Error: Instantiation failed: Use Actions.getInstance() instead of new.'
-      );
-    }
-
     logger.info("🤖 Beep bop - Creating the bot's account");
 
     try {
@@ -45,24 +38,21 @@ export default class Actions {
       waitReady().then(() => {
         this.account = keyring.addFromMnemonic(mnemonic);
 
-        this.updateFaucetBalance().then(() =>
-          logger.info('Fetched faucet balance 💰')
-        );
         // We do want the following to just start and run
         // eslint-disable-next-line @typescript-eslint/no-floating-promises
         // TODO: Adding a subscription would be better but the server supports on http for now
-        setInterval(() => this.updateFaucetBalance(), balancePollIntervalMs);
+        const updateFaucetBalance = (log = false) => {
+          if (log) logger.info('Fetched faucet balance 💰');
+          this.updateFaucetBalance().then(() =>
+            setTimeout(updateFaucetBalance, balancePollIntervalMs)
+          );
+        };
+        updateFaucetBalance(true);
       });
     } catch (error) {
       logger.error(error);
       errorCounter.plusOne('other');
     }
-
-    Actions.instance = this;
-  }
-
-  public static getInstance(): Actions {
-    return Actions.instance;
   }
 
   /**
@@ -256,3 +246,5 @@ export default class Actions {
     }
   }
 }
+
+export default new Actions();
--- a/src/server/services/Actions.ts
+++ b/src/server/services/Actions.ts
@@ -46,7 +46,7 @@ export default class Actions {
         this.account = keyring.addFromMnemonic(mnemonic);
 
         this.updateFaucetBalance().then(() =>
-          logger.info('Fetched faucet balance 🏦')
+          logger.info('Fetched faucet balance 💰')
         );
         // We do want the following to just start and run
         // eslint-disable-next-line @typescript-eslint/no-floating-promises
@@ -223,7 +223,7 @@ export default class Actions {
 
     if (isDripSuccessResponse(result)) {
       await this.updateFaucetBalance().then(() =>
-        logger.info('Updating balance 💰')
+        logger.info('Updated balance 💰')
       );
     }
 
--- a/src/server/services/Actions.ts
+++ b/src/server/services/Actions.ts
@@ -39,13 +39,12 @@ class Actions {
         this.account = keyring.addFromMnemonic(mnemonic);
 
         // We do want the following to just start and run
-        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         // TODO: Adding a subscription would be better but the server supports on http for now
         const updateFaucetBalance = (log = false) => {
-          if (log) logger.info('Fetched faucet balance 💰');
-          this.updateFaucetBalance().then(() =>
-            setTimeout(updateFaucetBalance, balancePollIntervalMs)
-          );
+          this.updateFaucetBalance().then(() => {
+            if (log) logger.info('Fetched faucet balance 💰');
+            setTimeout(updateFaucetBalance, balancePollIntervalMs);
+          });
         };
         updateFaucetBalance(true);
       });
@@ -184,13 +183,16 @@ class Actions {
     try {
       if (!this.account) throw new Error('account not ready');
 
-      if (faucetBalance && parsedAmount >= faucetBalance) {
+      if (
+        typeof faucetBalance !== 'undefined' &&
+        parsedAmount >= faucetBalance
+      ) {
         throw new Error(
           `Can't send "${parsedAmount}", as balance is smaller "${faucetBalance}"`
         );
       }
-      // start a counter and log a timeout error if we didn't get an answer in time
       const dripAmount = parsedAmount * 10 ** decimals;
+      // start a counter and log a timeout error if we didn't get an answer in time
       dripTimeout = rpcTimeout('drip');
       if (parachain_id != '') {
         result = await this.teleportTokens(dripAmount, address, parachain_id);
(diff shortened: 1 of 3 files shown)


chevdor (changes requested):
All good overall.

If I remember right, we do return failures/errors metrics to prometheus.
It would be nice to increment a new extrinsic failure counter so we can also monitor this from prometheus.

src/server/services/Actions.ts
```diff
@@ -25,10 +26,17 @@ const rpcTimeout = (service: string) => {
...
 
 export default class Actions {
+  private static instance: Actions = new Actions();
   account: KeyringPair | undefined;
   #faucetBalance: number | undefined;
 
   constructor() {
+    if (Actions.instance) {
+      throw new Error(
+        'Error: Instantiation failed: Use Actions.getInstance() instead of new.'
```
joao-paulo-parity, line 36: I think `export const actions = new Action()` and making `Action` unexported would be better than this. It would avoid the runtime check and make it less error-prone overall.

src/server/services/Actions.ts
```diff
@@ -37,35 +45,51 @@ export default class Actions {
...
-          // TODO: Adding a subscription would be better but the server supports on http for now
-          this.updateFaucetBalance().catch(console.error);
-        }, balancePollIntervalMs);
+        this.updateFaucetBalance().then(() =>
+          logger.info('Fetched faucet balance 💰')
+        );
+        // We do want the following to just start and run
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        // TODO: Adding a subscription would be better but the server supports on http for now
+        setInterval(() => this.updateFaucetBalance(), balancePollIntervalMs);
```
joao-paulo-parity, line 54: `setInterval` isn't a good idea because it can end up spamming the server with "stacked" requests if they end up overlapping (e.g. due to delays). It's better to reschedule the call only when the previous `updateFaucetBalance` finishes, i.e.

```js
const updateFaucetBalance = () => {
  this.updateFaucetBalance().then(() =>
    setTimeout(updateFaucetBalance, balancePollIntervalMs),
  )
}
updateFaucetBalance()
```

joao-paulo-parity (changes requested):

src/server/services/Actions.ts
mordamax: Good point, thank you!

mordamax:
> If I remember right, we do return failures/errors metrics to prometheus.
It would be nice to increment a new extrinsic failure counter so we can also monitor this from prometheus.

@chevdor yes, whenever this error happens, it does increase the error count there in catch block https://github.com/paritytech/substrate-matrix-faucet/blob/d8621996b84cc17370c51554fda3975df80161c4/src/server/services/Actions.ts#L218, as this balance check throws a new Error.

src/server/services/Actions.ts
```diff
@@ -45,24 +38,21 @@ export default class Actions {
...
         this.account = keyring.addFromMnemonic(mnemonic);
 
-        this.updateFaucetBalance().then(() =>
-          logger.info('Fetched faucet balance 💰')
-        );
         // We do want the following to just start and run
         // eslint-disable-next-line @typescript-eslint/no-floating-promises
         // TODO: Adding a subscription would be better but the server supports on http for now
-        setInterval(() => this.updateFaucetBalance(), balancePollIntervalMs);
+        const updateFaucetBalance = (log = false) => {
```
joao-paulo-parity, line 44: why have `log = false` if the only usage is `updateFaucetBalance(true)`?

src/server/services/Actions.ts
```diff
@@ -45,24 +38,21 @@ export default class Actions {
       waitReady().then(() => {
         this.account = keyring.addFromMnemonic(mnemonic);
 
-        this.updateFaucetBalance().then(() =>
-          logger.info('Fetched faucet balance 💰')
-        );
         // We do want the following to just start and run
         // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
joao-paulo-parity, line 42: Does this matter anymore?

src/server/services/Actions.ts
```diff
@@ -45,24 +38,21 @@ export default class Actions {
       waitReady().then(() => {
         this.account = keyring.addFromMnemonic(mnemonic);
 
-        this.updateFaucetBalance().then(() =>
-          logger.info('Fetched faucet balance 💰')
-        );
         // We do want the following to just start and run
```
joao-paulo-parity, line 41: Can this comment be removed?

src/server/services/Actions.ts
```diff
@@ -164,15 +178,20 @@ export default class Actions {
...
     let result: DripResponse;
+    const parsedAmount = Number(amount);
+    const faucetBalance = this.getFaucetBalance();
 
     try {
       if (!this.account) throw new Error('account not ready');
 
-      const dripAmount = Number(amount) * 10 ** decimals;
-
+      if (faucetBalance && parsedAmount >= faucetBalance) {
```
joao-paulo-parity, line 187: I think it should just be `parsedAmount >= faucetBalance`, otherwise it won't be caught when `faucetBalance` is `0`

src/server/services/Actions.ts
```diff
@@ -37,12 +38,16 @@ export default class Actions {
...
-          // We do want the following to just start and run
-          // eslint-disable-next-line @typescript-eslint/no-floating-promises
-          // TODO: Adding a subscription would be better but the server supports on http for now
-          this.updateFaucetBalance().catch(console.error);
-        }, balancePollIntervalMs);
+        // We do want the following to just start and run
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        // TODO: Adding a subscription would be better but the server supports on http for now
+        const updateFaucetBalance = (log = false) => {
+          if (log) logger.info('Fetched faucet balance 💰');
```
joao-paulo-parity, line 45: This log line is misleading because it says "Fetched" before the call is made. What about "Fetch`ing`"?

src/server/services/Actions.ts
```diff
@@ -164,15 +178,20 @@ export default class Actions {
...
       if (!this.account) throw new Error('account not ready');
 
-      const dripAmount = Number(amount) * 10 ** decimals;
-
+      if (faucetBalance && parsedAmount >= faucetBalance) {
+        throw new Error(
+          `Can't send "${parsedAmount}", as balance is smaller "${faucetBalance}"`
+        );
+      }
       // start a counter and log a timeout error if we didn't get an answer in time
```
joao-paulo-parity, line 192: This comment should be moved above `dripTimeout` I presume

src/server/services/Actions.ts
```diff
@@ -192,6 +211,12 @@ export default class Actions {
     // we got and answer reset the timeout
     if (dripTimeout) clearTimeout(dripTimeout);
 
+    if (isDripSuccessResponse(result)) {
+      await this.updateFaucetBalance().then(() =>
+        logger.info('Updated balance 💰')
```
joao-paulo-parity, line 216: I think you could move this logging to `updateFaucetBalance` directly

Also it should be more specific: `Updated faucet balance`. "Updated balance" doesn't tell me what kind of balance it updated.

src/server/services/Actions.ts
```diff
@@ -54,18 +59,27 @@ export default class Actions {
...
+      const { data: balances } = await polkadotApi.query.system.account(
+        this.account.address
+      );
+      const precision = 5;
+      this.#faucetBalance =
+        balances.free
+          .toBn()
+          .div(new BN(10 ** (decimals - precision)))
+          .toNumber() /
+        10 ** precision;
```
joao-paulo-parity, line 78: After this line is where it would be better to put the `Updated faucet balance`

joao-paulo-parity (changes requested):

src/server/services/Actions.ts
mordamax: To avoid polluting the logs with timeouted updates later, so it only logs first time on deploy and server run

src/server/services/Actions.ts
mordamax: Didn't want to move the log to actual `updateFaucetBalance` as it happens pretty often and may potentially pollute logs. I'll update wording though, you're right

src/server/services/Actions.ts
mordamax: I guess makes sense to remove after #82

src/server/services/Actions.ts
mordamax: [c513bbb](https://github.com/paritytech/substrate-matrix-faucet/pull/147/commits/c513bbba76ffeddfc0a5022380e7179fc87e2d97)

src/server/services/Actions.ts
mordamax: removed [c513bbb](https://github.com/paritytech/substrate-matrix-faucet/pull/147/commits/c513bbba76ffeddfc0a5022380e7179fc87e2d97)

joao-paulo-parity (approved):

chevdor (approved):

chevdor:
ref https://github.com/paritytech/substrate-matrix-faucet/issues/149

PR #150, opened 2022-04-25
Fix drips arbitrary amounts

### LINKED ISSUE #117: Drips for arbitrary amounts for whitelisted accounts are no longer working
This breaks some fairly "critical" functionality of the faucet which allows self-service of arbitrary amounts of tokens to whitelisted accounts. See below screenshot for the error.

2022-01-20-112632_2578x307_scrot

mordamax:
Fix #117 

Test evidence: 

Before:
image

After fix:
image


src/server/services/Actions.ts
```diff
@@ -191,7 +191,9 @@ class Actions {
           `Can't send "${parsedAmount}", as balance is smaller "${faucetBalance}"`
         );
       }
-      const dripAmount = parsedAmount * 10 ** decimals;
+
+      const dripAmount = BigInt(parsedAmount * 10 ** decimals);
```
chevdor, line 195: This is not a valid fix:
If `parsedAmount * 10 ** decimals` > `9007199254740991` you will run into the issue again.

You should take the drip amount from the user as string (so that `9999999999999999` is also valid), make it a BigInt, then do the math using the BigInt lib:
- https://github.com/substack/node-bigint#muln
- https://github.com/substack/node-bigint#mpown

Note that you need to be vary with the conversion: if a user passes 1.42, you do not want to convert that to a BigInt first (that would become `1`) but first apply the network decimals to make it a BigInt such as `1420000000` in planks.

chevdor (changes requested):
This fix is not valid

chevdor (approved):

mordamax:
@chevdor had to add tests here, please check once more

.gitlab-ci.yml
```diff
@@ -77,29 +77,23 @@ variables:
...
-    - yarn typing
-
 check-linting:
   stage:                           test
   <<:                              *common-refs
   <<:                              *kubernetes-env
   script:
-    - yarn --immutable --network-concurrency 1
-    - yarn lint
+    - apk add --no-cache python3 openssl make g++
```
alvicsam, line 85: Is it really necessary to install these packages?

alvicsam: Checked previous pipelines, I see why `python3` is needed. I can assume that `openssl` needs for ssl connections. But still can't understand about `make` and `g++`

src/utils.ts
```diff
@@ -8,3 +16,10 @@ export function isAccountPrivileged(sender: string): boolean {
     sender.endsWith(':matrix.parity.io') || sender.endsWith(':web3.foundation')
   );
 }
+
+export function convertAmountToBn(amount: string): bigint {
+  const parsedAmount = Number(amount);
+  return parsedAmount < MAX_ALLOWED_DRIP_FLOAT
+    ? BigInt(parsedAmount * 10 ** decimals)
```
joao-paulo-parity, line 23: I think it would be better to convert the amount to BigInt first and then do the multiplications after. e.g. `parsedAmount * 10 ** decimals` might overflow [`MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) for some reason.

i.e.

```typescript
  const parsedAmount = BigInt(amount);
  // ...
```

tsconfig.json
```diff
@@ -3,10 +3,10 @@
       "paths": { "matrix-js-sdk": ["node_modules/@types/matrix-js-sdk/index.d.ts"] },
       /* Basic Options */
       // "incremental": true,                   /* Enable incremental compilation */
-      "target": "es2017",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
+      "target": "es2020",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
```
joao-paulo-parity, line 6: It's better to tell which Node.js version you're targetting with this in the comments

https://github.com/paritytech/try-runtime-bot/blob/28cf9a41157543fe0fe0a6dcbc0b147f95f4aab8/tsconfig.json#L11

tsconfig.json
```diff
@@ -3,10 +3,10 @@
...
       "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
       "lib": [
-        "es2017",
+        "es2020",
```
joao-paulo-parity, line 9: It's better to tell which Node.js version you're targetting with this in the comments

https://github.com/paritytech/try-runtime-bot/blob/28cf9a41157543fe0fe0a6dcbc0b147f95f4aab8/tsconfig.json#L11

src/utils.ts
```diff
@@ -8,3 +16,10 @@ export function isAccountPrivileged(sender: string): boolean {
...
+    : BigInt(new bignum(amount).mul(bignum.pow(10, decimals)).toString());
```
joao-paulo-parity, line 24: Could you add a comment explaining why this `bignum` library is needed? I don't know why you can't just use the `BigInt` directly.

src/utils.ts
```diff
@@ -1,5 +1,13 @@
+import bignum from 'bignum';
+import { ConfigManager } from 'confmgr/lib';
 import log4js from 'log4js';
 
+import { faucetConfig } from './faucetConfig';
+
+const config = faucetConfig('server');
+const decimals = config.Get('BACKEND', 'NETWORK_DECIMALS') as number;
+const MAX_ALLOWED_DRIP_FLOAT = 9999999.9;
```
joao-paulo-parity, line 9: Add an assertion `assert(Number.isSafeInteger(MAX_ALLOWED_DRIP_FLOAT))` for sanity's sake

src/utils.ts
```diff
@@ -1,5 +1,13 @@
+import bignum from 'bignum';
+import { ConfigManager } from 'confmgr/lib';
 import log4js from 'log4js';
 
+import { faucetConfig } from './faucetConfig';
+
+const config = faucetConfig('server');
+const decimals = config.Get('BACKEND', 'NETWORK_DECIMALS') as number;
```
joao-paulo-parity, line 8: Add an assertion `assert(Number.isSafeInteger(MAX_ALLOWED_DRIP_FLOAT))` for sanity's sake, otherwise multiplications might become bugged

src/utils.ts
joao-paulo-parity, line 9: Why can't this variable be named `maxAllowedDripFloat`? It's inconsistent like this

.gitlab-ci.yml
```diff
@@ -77,29 +77,23 @@ variables:
...
 
-check-formatting:
+check-tests:
   stage:                           test
   <<:                              *common-refs
   <<:                              *kubernetes-env
   script:
-    - yarn --immutable --network-concurrency 1
-    - yarn format
+    - apk add --no-cache python3 openssl make g++
```
joao-paulo-parity, line 94: Add a `TODO` for creating an image which already has those packages and create a ticket for it

src/utils.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+      expected:
+        '999999999999999999999999999999999999999999999999999999000000000000',
+    },
+    {
+      amount: '999999999999999999999999999999999999999999999999999999.7',
+      expected:
+        '999999999999999999999999999999999999999999999999999999000000000000',
+    },
+  ].forEach((t) =>
+    test(`convertAmountToBn ${t.amount} => ${t.expected.toString()}`, () => {
```
joao-paulo-parity, line 25: `expected.toString()` is redundant because all values are string already

src/server/services/Actions.ts
```diff
@@ -101,7 +101,7 @@ class Actions {
   }
 
   async teleportTokens(
-    dripAmount: number,
+    dripAmount: number | bigint,
```
joao-paulo-parity, line 104: The input here should be `bigint` only for more robustness. Should you do this `convertAmountToBn` in

https://github.com/paritytech/substrate-matrix-faucet/blob/82728e9e91e1978dbd623c616fe8e8c53dd8c7a7/src/server/services/Actions.ts#L194

instead?

joao-paulo-parity (changes requested):

.gitlab-ci.yml
joao-paulo-parity: I guess they're for https://github.com/justmoon/node-bignum but I don't understand why this package is necessary, see my comment in https://github.com/paritytech/substrate-matrix-faucet/pull/150#discussion_r860737803.

https://github.com/justmoon/node-bignum/issues/114#issue-356377396 hints at the fact that we should probably use only the native `BigInt`, but I honestly haven't worked with Node.js numeric computing to have a strong opinion about this. Waiting to hear back from @mordamax.

mordamax: @alvicsam openssl is dep of node-bignum. I was adding them one by one, starting from openssl, then python3, it was still failing cuz wasn't able to find `make` etc.. however I tested locally with `gitlab-runner exec docker <command>` on my M1 chip, may be that could be a reason. It could be that it should work only with `openssl python3`, but on my machine this will still fail. Is there any issue with this to be added (so I better understand going forward)?

mordamax: @joao-paulo-parity re. numeric computing - honestly I was trying to avoid this dep as much as possible and tried tons of ways (that's why it was easier to make unit tests, so it's easier to play with it)
Here's the result of when I use only BigInt and direct multiplications

image

src/utils.ts
mordamax: I was thinking about it, and played a lot with different ways before had to add bignum dependency. See my comment with screenshot of how it ends up to work with JS :(

src/utils.ts
joao-paulo-parity: Got it. I mean the naming convention is inconsistent with other constants. For me it's not a big deal to leave this as it is.

mordamax: I see what you mean. What you tell is `const DECIMALS` and other similar would need to look also UpperSnakecase to be consistent? ) yea, that's true

src/utils.ts
joao-paulo-parity: My previous comment was wrong, sorry. I corrected it now.

This was a suggestion to catch errors where the value is wrong for some reason. Try the following:

```
const decimals = 100000000000000000000000000000000000000000000000
Number.isSafeInteger(decimals) // false
```

If you pass too big of a number which can't be represented as an integer, it'll become unsafe. So this is just an extra safety measure to ensure we don't accidentally use an invalid value here. I don't know how big of a value you're expecting for this variable.

src/utils.ts
joao-paulo-parity: This suggestion actually doesn't work because Number.isSafeInteger doesn't work for floats, sorry

src/utils.ts
mordamax: `decimals` is also constant, (we certainly need to make naming more consistent), so also won't change by user. 
The only things will be user-inputed - is `amount: string`, which (per [conversation](https://github.com/paritytech/substrate-matrix-faucet/pull/150#discussion_r858509240) with @chevdor) could be big enough to produce number which is bigger than safe integer, that's why we use bignum also

.gitlab-ci.yml
alvicsam:
```suggestion
    # TODO: remove after https://github.com/paritytech/ci_cd/issues/406
    - apk add --no-cache python3 openssl make g++
```

Okay, it's just better to keep minimum of additional packages in CI image. Also it seems that it's time to create a CI image for `substrate-matrix-faucet` as @joao-paulo-parity suggested.

.gitlab-ci.yml
mordamax: [Added](https://github.com/paritytech/substrate-matrix-faucet/pull/150/files#diff-037ea159eb0a7cb0ac23b851e66bee30fb838ee8d0d99fa331a1ba65283d37f7R85)

.gitlab-ci.yml
```diff
@@ -77,29 +77,23 @@ variables:
...
```
alvicsam, line 94:
```suggestion
    # TODO: remove after https://github.com/paritytech/ci_cd/issues/406
    - apk add --no-cache python3 openssl make g++
```

.gitlab-ci.yml
joao-paulo-parity: For the record I briefly tried https://github.com/davidmartinez10/bigfloat which is pure Node.js and it gave some different results

```
> yarn test
yarn run v1.22.18
$ jest
 FAIL  src/utils.spec.ts
  utils
    ✓ convertAmountToBn 0 => 0 (2 ms)
    ✓ convertAmountToBn 1 => 1000000000000 (1 ms)
    ✓ convertAmountToBn 100 => 100000000000000
    ✓ convertAmountToBn 1.45 => 1450000000000 (1 ms)
    ✓ convertAmountToBn 145.454141111 => 145454141111000
    ✕ convertAmountToBn 99999999.9 => 99999999000000000000 (2 ms)
    ✓ convertAmountToBn 999999999 => 999999999000000000000
    ✕ convertAmountToBn 999999999.9 => 999999999000000000000
    ✕ convertAmountToBn 999999999999999.9 => 999999999999999000000000000
    ✓ convertAmountToBn 999999999999999999999999999999999999999999999999999999 => 999999999999999999999999999999999999999999999999999999000000000000
    ✕ convertAmountToBn 999999999999999999999999999999999999999999999999999999.7 => 999999999999999999999999999999999999999999999999999999000000000000 (1 ms)

  ● utils › convertAmountToBn 99999999.9 => 99999999000000000000

    expect(received).toEqual(expected) // deep equality

    Expected: "99999999000000000000"
    Received: "99999999900000000000"

      24 |   ].forEach((t) =>
      25 |     test(`convertAmountToBn ${t.amount} => ${t.expected}`, () => {
    > 26 |       expect(convertAmountToBn(t.amount).toString()).toEqual(t.expected);
         |                                                      ^
      27 |     })
      28 |   );
      29 | });

      at Object.<anonymous> (src/utils.spec.ts:26:54)

  ● utils › convertAmountToBn 999999999.9 => 999999999000000000000

    expect(received).toEqual(expected) // deep equality

    Expected: "999999999000000000000"
    Received: "999999999900000000000"

      24 |   ].forEach((t) =>
      25 |     test(`convertAmountToBn ${t.amount} => ${t.expected}`, () => {
    > 26 |       expect(convertAmountToBn(t.amount).toString()).toEqual(t.expected);
         |                                                      ^
      27 |     })
      28 |   );
      29 | });

      at Object.<anonymous> (src/utils.spec.ts:26:54)

  ● utils › convertAmountToBn 999999999999999.9 => 999999999999999000000000000

    expect(received).toEqual(expected) // deep equality

    Expected: "999999999999999000000000000"
    Received: "999999999999999900000000000"

      24 |   ].forEach((t) =>
      25 |     test(`convertAmountToBn ${t.amount} => ${t.expected}`, () => {
    > 26 |       expect(convertAmountToBn(t.amount).toString()).toEqual(t.expected);
         |                                                      ^
      27 |     })
      28 |   );
      29 | });

      at Object.<anonymous> (src/utils.spec.ts:26:54)

  ● utils › convertAmountToBn 999999999999999999999999999999999999999999999999999999.7 => 999999999999999999999999999999999999999999999999999999000000000000

    expect(received).toEqual(expected) // deep equality

    Expected: "999999999999999999999999999999999999999999999999999999000000000000"
    Received: "999999999999999999999999999999999999999999999999999999700000000000"

      24 |   ].forEach((t) =>
      25 |     test(`convertAmountToBn ${t.amount} => ${t.expected}`, () => {
    > 26 |       expect(convertAmountToBn(t.amount).toString()).toEqual(t.expected);
         |                                                      ^
      27 |     })
      28 |   );
      29 | });

      at Object.<anonymous> (src/utils.spec.ts:26:54)

Test Suites: 1 failed, 1 total
Tests:       4 failed, 7 passed, 11 total
Snapshots:   0 total
Time:        2.236 s, estimated 3 s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

```
import { BigFloat } from 'bigfloat.js'

export function convertAmountToBn(amount: string): BigFloat {
  const parsedAmount = new BigFloat(amount)

  return parsedAmount.mul(new BigFloat(10).pow(new BigFloat(decimals)))
}
```

What do you think @mordamax @chevdor?

.gitlab-ci.yml
```diff
@@ -77,29 +77,23 @@ variables:
...
-    - yarn typing
-
 check-linting:
   stage:                           test
   <<:                              *common-refs
   <<:                              *kubernetes-env
   script:
-    - yarn --immutable --network-concurrency 1
-    - yarn lint
+    - apk add --no-cache python3 openssl make g++ # TODO: replace this with image which includes these deps
```
alvicsam, line 85:
```suggestion
    # TODO: remove after https://github.com/paritytech/ci_cd/issues/406
    - apk add --no-cache python3 openssl make g++
```

src/utils.ts
joao-paulo-parity: > other similar would need to look also UpperSnakecase to be consistent

I would *prefer* the other way around: all variables should be `camelCase`. Regardless, I think consistency is what's important, regardless of preference.

.gitlab-ci.yml
mordamax: This looks nicer for sure :)  🥇
removes this ugly constant and too long comment

chevdor: It may be worth staying in line with the PolkadotJS API and use https://github.com/polkadot-js/common/tree/master/packages/x-bigint for the BigInt. Do you mind testing it ?
It is the one used by the API already so it will both limit the deps and stay in sync with the API.

mordamax: @chevdor I did test it, it's similar to regular BigInt, it seems more like a shim

mordamax: with BigFloat.js we were able 
- get rid of  ugly logic around max_float
- remove unnecessary openssl dependency and need to create an image
- add support of float decimals on big numbers

so It looks to be as preferable solution

chevdor: Ok, those are indeed sweet benefits for our use case but I don't see it very maintained with no change in 2 years.
If we go this route, we may want to depend on a fork we make so we can at least slide it some updates if required and try to get them upstream ?

# npm audit report

ansi-regex  4.0.0 - 4.1.0 || 5.0.0
Severity: high
 Inefficient Regular Expression Complexity in chalk/ansi-regex - https://github.com/advisories/GHSA-93q8-gq69-wqmw
 Inefficient Regular Expression Complexity in chalk/ansi-regex - https://github.com/advisories/GHSA-93q8-gq69-wqmw
fix available via `npm audit fix`
node_modules/ansi-regex
node_modules/string-width/node_modules/ansi-regex

glob-parent  <5.1.2
Severity: high
Regular expression denial of service in glob-parent - https://github.com/advisories/GHSA-ww39-953v-wcq6
fix available via `npm audit fix`
node_modules/glob-parent

lodash  <4.17.21
Severity: high
Command Injection in lodash - https://github.com/advisories/GHSA-35jh-r3h4-6jhm
fix available via `npm audit fix`
node_modules/lodash

minimist  <1.2.6
Severity: critical
Prototype Pollution in minimist - https://github.com/advisories/GHSA-xvch-5gv4-984h
fix available via `npm audit fix`
node_modules/minimist

4 vulnerabilities (3 high, 1 critical)

All of them can be fixed with `npm audit fix`.

mordamax: @chevdor good point, yea i think it makes sense, as current owner doesn't seem to be responsive there
do you think it makes sense do now? like creating a new issue for doing this

.gitlab-ci.yml
```diff
@@ -77,29 +77,21 @@ variables:
...
-
 check-linting:
   stage:                           test
   <<:                              *common-refs
   <<:                              *kubernetes-env
   script:
-    - yarn --immutable --network-concurrency 1
-    - yarn lint
+    - yarn --immutable
+    - yarn typing && yarn format && yarn lint
```
joao-paulo-parity, line 86: I was confused of what this `typing` command was supposed to be. Can't it be renamed to `typecheck`? That's how it's called in other OpsTooling projects

joao-paulo-parity (approved):

.gitlab-ci.yml
chevdor: I think we should. For a prod system with value I would say this is a must, but here for the faucet, we can make it easier on ourselves. It is always better to depend on the most maintained repo and if there is no such a repo, better use one of our own were we at least can patch things quickly as needed. If we patch the repo, we should obviously do a PR upsteam and **hope** the owner will take it. If (s)he does, great, if not, we know that the repo should not be counted on.

chevdor: I will approve in the current state but please let's track that down so we don't forget.

PR #153, opened 2022-04-26
split server and bot configs

### LINKED ISSUE #142: Revamp confmgr validation
During #87 implementation I had to disable validation, as initially confmgr is integrated as single app and 2 modules, and validation is running through whole application. 

The problem with that is that Bot and Server are actually 2 different applications, and they run with different set of ENV variables, which leads to failed validation, when Bot is running w/o Server's config and vice versa

Solution: 
- Split Bot and Server yml configs, initialization and validations, so that when each app starts - it validates only its own env variables
- Exit(1) if validation didn't pass

mordamax:
Fix #142


joao-paulo-parity (approved):

chevdor:
To add to the previous as you may have skipped that part:
https://gitlab.com/chevdor/confmgr/-/blob/master/samples/03_ts_yaml/configSpecs.yml

confmgr allows defining an App (prefix) that has several `modules`. Each module having its own set of config keys.

mordamax:
> To add to the previous as you may have skipped that part: https://gitlab.com/chevdor/confmgr/-/blob/master/samples/03_ts_yaml/configSpecs.yml
> 
> confmgr allows defining an App (prefix) that has several `modules`. Each module having its own set of config keys.

Sorry, I forgot to attach "Fix #142" where i have the explanation for that.
Basically each build goes with only Bot or Server set of envs, and if you run one file, it validates whole file, and it fails either for bot or server, cuz the according part of envs is missing

chevdor:
Ahh there we go, indeed, now it makes sense.

chevdor (approved):

PR #159, opened 2022-05-12
Automate rollbacks on failed upgrades

PierreBesson:
Also remove the dry-run part as it is useless if you actually upgrade just after.


.gitlab-ci.yml
```diff
@@ -136,19 +136,10 @@ publish-docker-server:
   script:
     - echo ${KUBE_NAMESPACE}
     - echo ${CI_ENVIRONMENT_NAME}
-    # do a dry-run 1st
-    - helm upgrade ${CI_ENVIRONMENT_NAME} kubernetes/faucetbot
```
mordamax, line 140: why do we need to remove this?

PR #176, opened 2022-10-06
Trying to fix prod deployment

mordamax:


PierreBesson:
No it's not correct. The faucet channels are on matrix.org.

PR #178, opened 2022-10-10
Added deployment for wococo network.

BulatSaif:


mordamax (approved):

BulatSaif:
We will use argocd for deployment

PR #185, opened 2023-01-25
Finish the request handling in the case of an error

rzadp:
Just a tiny change that finishes handling the request in the event of an error.
Otherwise, the request hangs forever, as tested locally with `curl`.

### DIFF
--- a/src/server/routes/actions.ts
+++ b/src/server/routes/actions.ts
@@ -19,6 +19,7 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
     .catch((e) => {
       logger.error(e);
       errorCounter.plusOne("other");
+      res.send({ balance: "0" });
     });
 });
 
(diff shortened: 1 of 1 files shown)


src/server/routes/actions.ts
```diff
@@ -55,6 +56,7 @@ router.post<unknown, DripResponse, BotRequestType>("/bot-endpoint", (req, res) =
     .catch((e) => {
       logger.error(e);
       errorCounter.plusOne("other");
+      res.send({ error: "Operation failed." });
```
rzadp, line 59: Not disclosing any more information to the external user because we don't know what `e` is, and worst case it could contain some sensitive info.

src/server/routes/actions.ts
```diff
@@ -19,6 +19,7 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
     .catch((e) => {
       logger.error(e);
       errorCounter.plusOne("other");
+      res.send({ balance: "0" });
```
rzadp, line 22: This endpoint doesn't have an error type of result, so following the existing pattern at the moment.

mutantcornholio (approved):

PR #188, opened 2023-01-26
Propose a new, externally accessible faucet endpoint

rzadp:
This PR proposes a new `/drip` endpoint, which could be externally accessible for the users, as requested by https://github.com/paritytech/opstooling/issues/172

It introduces a new configuration option `EXTERNAL_ACCESS`. If it's disabled, there should be no changes to the behaviour of the backend. It is disabled by default.

If it's enabled, it creates a new endpoint which is rate-limited by IP address of the requester.

The PR also introduces a new `docker-compose` file which can be used to spin up a new, seperate faucet with separate funds. That way, we require no changes to existing faucets - they can continue to not expose any ports and any risks would be limited to this new deployment of the faucet.

### What would be next

If we decide to continue going this way, I would be adding changes such us [request sanitization](https://github.com/paritytech/substrate-matrix-faucet/pull/187) or limiting daily total limit of faucet payouts.
In this PR I tried to make minimum viable changes.

Also I would think about resetting the IP limit more frequently, e.g. once per hour. Because if two developers would be trying to drip on the same wifi they wouldn't be able to for the whole day.

### DIFF
--- a/src/server/routes/actions.ts
+++ b/src/server/routes/actions.ts
@@ -4,6 +4,7 @@ import { isDripSuccessResponse } from "../../guards";
 import { logger } from "../../logger";
 import { BalanceResponse, DripRequestType, DripResponse } from "../../types";
 import { isAccountPrivileged } from "../../utils";
+import { config } from "../config";
 import { metricsDefinition } from "../constants";
 import actions from "../services/Actions";
 import ActionStorage from "../services/ActionStorage";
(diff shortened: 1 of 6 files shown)


src/server/routes/actions.ts
```diff
@@ -52,14 +53,37 @@ const dripRequestHandler = async (requestOpts: DripRequestType): Promise<DripRes
   }
 };
 
-router.post<unknown, DripResponse, DripRequestType>("/bot-endpoint", (req, res) => {
-  dripRequestHandler(req.body)
-    .then(res.send)
-    .catch((e) => {
+if (config.Get("EXTERNAL_ACCESS")) {
+  logger.info("The faucet is externally accessible for drip requests.");
+  router.post<unknown, DripResponse, DripRequestType>("/drip", async (req, res) => {
```
mordamax, line 58: why not having one endpoint for both bot & web?
then inside of the handler, you'd have clear way to validate, depends on EXTERNAL_ACCESS, it's either `false` 1) username 2) address, or `true` 1) IP 2) address
currently, when we use 2 separate api endpoints, the validation is a bit spread b/w the api controller and dripRequestHandler. Ideally, if we use dripRequestHandler in both apis, they should accept 2 clear contracts, depends on the EXTERNAL_ACCESS state.

src/server/services/ActionStorage.ts
```diff
@@ -32,23 +33,34 @@ export default class ActionStorage {
...
 
-  async isValid(username: string, addr: string): Promise<boolean> {
-    username = sha256(username);
-    addr = sha256(addr);
-
-    const totalUsername = await this.query(username);
-    const totalAddr = await this.query(addr);
-
-    return Number(totalUsername) < LIMIT_USERS && Number(totalAddr) < LIMIT_ADDRESSES;
+  async isValid(opts: { username?: string; addr?: string; ip?: string }): Promise<boolean> {
```
mordamax, line 36: it looks a bit ambiguous now to me. if nothing is provided, then it's valid?

here https://github.com/paritytech/substrate-matrix-faucet/pull/188/files#diff-ad701bbca301460f45616bd9f07dd4cca0c72c3ed856701d5e9ed9ebee910100R31 if sender & address wasn't provided, it will be `isValid === true`?

mordamax (changes requested):
I'd suggest to either make 1 api and move clear validation between the EXTERNAL_ACCESS state, or, if 2 apis, then more clear validation rules of IP or Username depends on EXTERNAL_ACCESS

rzadp:
Closed in favor of https://github.com/paritytech/substrate-matrix-faucet/pull/192

Issue #190, opened 2023-01-31
TypeError in internal faucet deployment


PierreBesson:
In our internal faucet deployment for the Wococo network. We are seeing the following server error log:

```
[2023-01-31T11:58:31.513] [ERROR] default - TypeError: Cannot read properties of undefined (reading 'req')
    at send (/backend/node_modules/express/lib/response.js:110:18)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```
Moreover, the bot reports `Operation failed.` in the chat room even though the transfer is successful.

Version: `paritytech/faucet-server:v1.4.0-1a78cb70`
Bot config:
```
SMF_BOT_NETWORK_DECIMALS: 12
SMF_BOT_NETWORK_UNIT: UNIT
SMF_BOT_DRIP_AMOUNT:1
SMF_BOT_FAUCET_IGNORE_LIST:
SMF_BOT_MATRIX_BOT_USER_ID: @wococo-faucet:matrix.org
SMF_BOT_MATRIX_SERVER: https://matrix.org
````

Server config:
```
SMF_BACKEND_INJECTED_TYPES: {}
SMF_BACKEND_NETWORK_DECIMALS: 12
SMF_BACKEND_PORT: 5555
SMF_BACKEND_RPC_ENDPOINT: https://http-wococo-rpc-node-0.parity-testnet.parity.io/
```

mordamax:
@PierreBesson sorry, i don't see the the wococo in our configs https://github.com/search?q=repo%3Aparitytech%2Fsubstrate-matrix-faucet+wococo&type=code

have no idea how that was set up

could you please share with us how was it installed?

PierreBesson:
I pasted configs in the issue.

mordamax:
yea sorry, i was able to reproduce this on all networks on main branch

looks like what we have now "on prod" (rococo & westend) is not impacted, so it has to be with few last PRs

since `v1.4.0-1a78cb70` works good, on `main-e1ced066` reproducible

@rzadp could you please take a look? ☝️

PR #191, opened 2023-02-01
Properly bind response sending function

rzadp:
Closes https://github.com/paritytech/substrate-matrix-faucet/issues/190

I have plans to introduce some E2E smoke tests to catch this kind of stuff.


Bullrich (approved):
Looks OK to me

PR #192, opened 2023-02-02
Add an option to handle external requests protected by ReCaptcha

rzadp:
Added a `SMF_BACKEND_EXTERNAL_ACCESS` variable.
If set, the drip endpoint will expect external, untrusted requests protected by recaptcha.

Added some tests for the request handling in both external and non-external mode.

Refactored the `ActionStorage` and `Actions` to be injected into the drip request handler to allow running the tests easily.

Progresses https://github.com/paritytech/opstooling/issues/172


rzadp:
How it can be used: https://github.com/paritytech/ink-docs/pull/145

src/faucetConfig.ts
```diff
@@ -4,7 +4,13 @@ import botConfigSpec from "../env.bot.config.json";
 import serverConfigSpec from "../env.server.config.json";
 import { logger } from "./logger";
 
-type SpecType<T> = T extends { type: "string" } ? string : T extends { type: "number" } ? number : never;
+type SpecType<T> = T extends { type: "string" }
+  ? string
+  : T extends { type: "number" }
+  ? number
```
mordamax, line 10: that should go with additional tab indentation?

src/server/routes/actions.ts
```diff
@@ -23,43 +26,49 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
...
+router.post<unknown, DripResponse, Partial<DripRequestType>>("/drip", async (req, res) => {
+  try {
+    const { address, parachain_id, amount, sender, recaptcha } = req.body;
+    if (!address) {
+      res.send({ error: "Missing parameter: 'address'" });
+      return;
     }
-
-    return sendTokensResult;
+    if (!parachain_id) {
```
mordamax, line 38: parachain_id is optional, so I guess we shouldn't return an error

rzadp:
Answering some unaddressed questions from https://github.com/paritytech/substrate-matrix-faucet/pull/188#pullrequestreview-1273138867

> do we need to disable probot if that's EXTERNAL_ACCESS?

Yes at the moment, until we recognize which requests are coming from a (trusted) bot.
I have an idea for that - we would check if the `req.ip` is an internal, reserved IP address. (In the case of docker siblings, it's `172.16.x.x`, at least on my machine)
I would be doing that separately.

> do we need to rename substrate-matrix-faucet to substrate-faucet if it starts to be more flexible?

I agree with that.

src/server/services/DripRequestHandler.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  /**
+   * The requests are coming from external requesters.
+   */
+  describe("With external access", () => {
+    const defaultRequest = {
+      external: true,
+      amount: "0.5",
+      parachain_id: "1002",
```
mordamax, line 96: tests are great!  however, since parachain_id is optional, should be cases with & without parachain id

src/server/services/Recaptcha.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { logger } from "../../logger";
+import { config } from "../config";
+import errorCounter from "./ErrorCounter";
+
+export class Recaptcha {
+  constructor(private secret: string = config.Get("RECAPTCHA_SECRET")) {}
+
+  async validate(captcha: string): Promise<boolean> {
+    if (!this.secret) {
```
mordamax, line 12: i guess ideally it should be checked during startup? so it's not running app, if the EXTERNAL === true and no recaptcha secret. Could be here too, but if you run application, all green 🌈 but users start sending you errors (and in this case it's `critical`s, as they won't be able to operate with core functionality) that's a bit more unfortunate case, compared to if app wouldn't start and would be automatically rolled-back to working version in kube

mordamax (changes requested):

src/server/services/Recaptcha.ts
rzadp: Agree, updated to crash on startup.

src/server/services/DripRequestHandler.spec.ts
rzadp: Added those cases.

src/server/routes/actions.ts
rzadp: I didn't realize that `parachain_id` can be empty - updated.

src/faucetConfig.ts
rzadp: It's auto-linted like that by our linting rules.

mordamax (approved):

PR #194, opened 2023-02-02
Introduce Rococo external faucet kubernetes config

rzadp:


mordamax (approved):

mordamax:
would also be great to introduce stg env for external:true, perhaps wococo or versi

mordamax:
> > would also be great to introduce stg env for external:true, perhaps wococo or versi
> 
> Could you point me again to the place where the staging is set up? I don't see any usage of `deploy-stg-refs` in the [repo](https://github.com/paritytech/substrate-matrix-faucet/blob/main/.gitlab-ci.yml), or any configuration related to wococo or versi. That is in a different repostiory, correct? Would we like the same for `rococo-external` staging?

sure -> https://github.com/paritytech/command-bot/tree/master/helm

Bullrich (approved):
LTM

PR #197, opened 2023-02-07
[Paused] Add Google Authentication to the external faucet

rzadp:
According to https://github.com/paritytech/opstooling/issues/172#issuecomment-1419461805

With google authentication, we can get a username (email) of the requester, so with this we can make it more consistent with the non-external mode by saving requester's name and limiting it.
And also we can make it consistent by allowing `@parity.io` users unlimited access.


rzadp:
The new example usage in ink docs is here: https://github.com/paritytech/ink-docs/pull/153

mutantcornholio (approved):

Bullrich (approved):
Seems fine by me but also brings me the question. At this point shouldn't we simply make a new faucet website captcha?

The matrix one was useful because matrix provided the authentication part. If are doing our own authentication doesn't it beat the whole process of using matrix?

CC: @mutantcornholio @mordamax

mordamax (changes requested):
the implementation is perfect, however, as we discussed on our team meeting - google SSO in the docs about decentralized technologies is kinda contradicts the whole idea :) 
Even though the security team recommends this as a way to reduce certain risks, we should've also agree this with our consumer first. By the comments in the thread it seems that indeed they are not very happy, so would be good to get on same page with most parties (us, Michi, security, devops) and then implement the final solution

PR #198, opened 2023-02-14
Introduce E2E tests for the faucet bot and server

rzadp:
First, we spin up:
- A local Polkadot node in a dev mode
- A local matrix server
  - Default configuration, except for increased rate limits
  - We create necessary users and gather them in a room
- The faucet backend
  - It uses `//Alice` seed so it is funded
- The faucet bot

Next, we test the behaviour by sending `!balance` and `!drip` messages.

### DIFF
--- a/.github/workflows/E2E.yml
+++ b/.github/workflows/E2E.yml
@@ -15,7 +15,7 @@ jobs:
         uses: actions/setup-node@v3.5.1
         with:
           node-version: "16.10"
-      - run: yarn install --frozen-lockfile
+      - run: yarn install --network-concurrency 1 --frozen-lockfile
       - name: Bootstrap Polkadot and Matrix
         run: |
           chmod -R o+rwx ./e2e/matrix_data
--- a/.github/workflows/E2E.yml
+++ b/.github/workflows/E2E.yml
@@ -1,0 +1,42 @@
+name: E2E
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+jobs:
+  e2e:
+    timeout-minutes: 15
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v3.3.0
+      - name: Setup Node.js for use with actions
+        uses: actions/setup-node@v3.5.1
+        with:
+          node-version: "16.10"
+      - run: yarn install --frozen-lockfile
+      - name: Bootstrap Polkadot and Matrix
+        run: |
+          chmod -R o+rwx ./e2e/matrix_data
+          ./e2e/bootstrap.sh
+      - name: Run the faucet backend
+        run: |
+          yarn dev:backend > backend.txt 2>&1 &
+          ./e2e/wait_until.sh 'curl -s "localhost:5555"'
+      - name: Run the faucet bot
+        run: yarn dev:bot > bot.txt 2>&1 &
+      - name: Run the E2E tests
+        run: yarn test:e2e
+      - name: Debug Polkadot logs
+        if: failure()
+        run: cd e2e && docker-compose logs polkadot
+      - name: Debug Matrix logs
+        if: failure()
+        run: cd e2e && docker-compose logs matrix
+      - name: Debug backend logs
+        if: failure()
+        run: cat backend.txt
+      - name: Debug bot logs
+        if: failure()
+        run: cat bot.txt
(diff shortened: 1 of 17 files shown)


.github/workflows/E2E.yml
```diff
@@ -0,0 +1,44 @@
...
+      - main
+jobs:
+  e2e:
+    timeout-minutes: 15
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v3.3.0
+      - name: Install prerequisites
+        run: sudo apt-get update && sudo apt-get install -y jq
```
Bullrich, line 15: `jq` is [already bundled in ubuntu-22.04](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md)

Bullrich (changes requested):
Looks to me.

But seriously, good pr. **Lots** of setups I see.

Should we also add tests about the bot interacting with invalid addresses?

.github/workflows/E2E.yml
rzadp: Thanks, good to know - removed the install.

rzadp:
> Should we also add tests about the bot interacting with invalid addresses?

Yes, I would be adding some more test cases in a separate PR.

mordamax (approved):

Bullrich (approved):

src/utils.ts
```diff
@@ -1,3 +1,20 @@
...
+  const stopTime = Date.now() + timeout;
+  while (Date.now() <= stopTime) {
+    const result = await predicate();
+    if (result) {
+      return result;
+    }
+    await sleep(step);
+  }
+  throw new Error(`waitUntil timed out after ${timeout} ms`);
+};
```
mutantcornholio, line 20: There are already `delay` and and `until` functions in `opstooling-js`

mutantcornholio (changes requested):
Small nit about reusing common code, still going through test setup configuration

src/utils.ts
```diff
@@ -1,3 +1,20 @@
 export function isAccountPrivileged(sender: string): boolean {
   return sender.endsWith(":parity.io") || sender.endsWith(":web3.foundation");
 }
+
+export const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Predicate = () => any | (() => Promise<any>);
+
+export const waitUntil = async (predicate: Predicate, step = 100, timeout = 10000) => {
```
mutantcornholio, line 10: Also, the timeout here's the same as the governing timeout for the tests, meaning that this timeout condition will never fire, and instead we'll be getting timeouts from jest, without stacktrace

PierreBesson:
IMO setting up docker containers using docker exec and wait-for in bash scripts can be flaky. I would recommend using a proper docker test framework such as https://github.com/testcontainers/testcontainers-node

mutantcornholio:
> IMO setting up docker containers using docker exec and wait-for in bash scripts can be flaky. I would recommend using a proper docker test framework such as https://github.com/testcontainers/testcontainers-node

Good point actually. This advice could save me quite some time a month ago :)

rzadp:
> IMO setting up docker containers using docker exec and wait-for in bash scripts can be flaky. I would recommend using a proper docker test framework such as https://github.com/testcontainers/testcontainers-node

I had no idea this existed, I will check it out.

Bullrich:
> IMO setting up docker containers using docker exec and wait-for in bash scripts can be flaky. I would recommend using a proper docker test framework such as https://github.com/testcontainers/testcontainers-node

That looks super useful! Will keep it in mind

mutantcornholio (approved):

sergejparity (approved):

src/utils.ts
rzadp: Thanks for pointing that out - applied.

src/utils.ts
rzadp: Good catch, timeouts adjusted.

e2e/bootstrap.sh
```diff
@@ -0,0 +1,67 @@
...
+# Clean up potential data from previous runs,
+# in order to start with a clean state.
+docker-compose down -v
+rm -rf ./matrix_data/homeserver.db* ../sqlite.db
+
+
+# Start Polkadot and Matrix and wait until they are up.
+docker-compose up -d
+source wait_until.sh 'curl -s "localhost:8008"'
+source wait_until.sh 'curl -s "localhost:9933"'
```
parutger, line 14: This should technically be 127.0.0.1 instead of localhost

rzadp: Ok, updated.

PR #204, opened 2023-02-20
Created client for the faucet

### LINKED ISSUE #203: Create Website faucet
Based on the approval of the [Proposal to create a new and more easy to use faucet](https://forum.parity.io/t/proposal-to-create-a-new-and-more-easy-to-use-faucet/) we have decided to move on with this project.

The MVP for the project will consist on a single website presenting three items:
- An input field for an address
- A captcha
- A submit button

Some UX is expected but I will leave that for the developer of the site (stuff like error handling, animations, etc.)

Bullrich:
Created a client for the faucet.

You can test the website in http://bullrich.dev/substrate-matrix-faucet/ 
Enter any address to try the happy path, and write `error` as the address to try see how it reacts under an error.

This commit closes #203

Created GitHub Action to deploy and test the page automatically.

We still need to enable an endpoint to work for this client, and to set up a new Google Captcha account.

This project is using Google Captcha v3 as it's more friendly and less invasive.

It was developed using [Svelte](https://svelte.dev), [Tailwind](https://tailwindcss.com/) and [DaisyUI](daisyui.com/).

I also added meta tags so when it's share it looks prettier.


.github/workflows/deploy-site.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Install node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 18
+      - run: npm ci
+      - run: npm run check
+      - run: npm run build
+        env:
+          VITE_CAPTCHA_KEY: 6LftP5wkAAAAAH5cbuutkYKKaTnUMTLPqWlyeIL6
+          VITE_FAUCET_URL: https://ink-docs-rococo-faucet.parity-testnet.parity.io/drip
```
mordamax, line 31: this is temporary right?
if that's gonna be an alternative to current westend/rococo in matrix, there should be separate instances of server with its own wallets inside. So there will be 2-3 deploy-site.yml jobs, each with its own VITE_FAUCET_URL?

mordamax: nit: perhaps would be a bit more flexible going forward to have this as host, and /drip or /balance would be defined in the query/request methods

client/README.md
```diff
@@ -0,0 +1,31 @@
+# Web Faucet
+
+Web Client to access the faucet. Powered by Catpcha v3
+
+## Why?
+
+We don't have any easy to use Faucet, our two current options are to [access Matrix and contact a bot](https://wiki.polkadot.network/docs/learn-DOT#getting-tokens-on-the-rococo-testnet) or look for [faucet available inside the Ink! documentation](https://use.ink/faucet).
```
mordamax, line 7:
```suggestion
To simplify the use the Faucet through a standalone webpage and open ability to share it via social networks & google search.
Two current options are to [access Matrix and contact a bot](https://wiki.polkadot.network/docs/learn-DOT#getting-tokens-on-the-rococo-testnet) or [Ink! documentation](https://use.ink/faucet).
```

mordamax (approved):

.github/workflows/deploy-site.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          name: faucet
+          path: ./client/dist
+          if-no-files-found: error
+  deploy-to-github-pages:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
```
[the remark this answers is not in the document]
Bullrich: Yes, we have two steps for this action. The first one tests and builds the client, and the second one publishes it to github pages.

I set it up so we can use the first part as a check for the PRs but stop it from deploying. With the current implementation, when it is merged that step will run again, but this time the deployment step won't be skipped, allowing it to be published.

.github/workflows/deploy-site.yml
Bullrich: Yes, I copy this from the implementation of !ink, but I need to sync with @rzadp so that we can see how we are gonna deploy it.

Also, this implementation uses captcha v3 (less invasive) while that endpoint uses captcha v2, so I think it also needs a slightly different code.

client/README.md
Bullrich: Thanks for the suggestion! Will apply it!

client/package-lock.json
```diff
@@ -0,0 +1,2063 @@
+{
```
rzadp, line 1: We use `yarn.lock` everywhere else, unless there is some reason for not using yarn here?

client/src/lib/utils/faucetRequest.ts
```diff
@@ -0,0 +1,39 @@
...
+    return boilerplateRequest(address, recaptcha);
+  }
+
+  return faucetRequest(address, recaptcha);
+}
+
+export async function faucetRequest(address: string, recaptcha: string): Promise<string> {
+  const body = {
+    address,
+    parachain_id: "1002",
```
rzadp, line 14: Shall we make this into a dropdown or something? (Maybe in next PR).
Even if we would only have 1 option right now, at least we would see where exactly the funds would go.

rzadp (approved):
Looks all good to me, except the minor comments that could be handled separately.

sergejparity (approved):

client/src/lib/utils/faucetRequest.ts
Bullrich: Yeah, I agree.

I'll merge the current progress and I'll make a new PR with the id as a dropdown, that way the complexity is smaller and we can also brainstorm a bit the idea on how to do it.

rzadp: > that way the complexity is smaller and we can also brainstorm a bit the idea on how to do it

100% agree.

PR #206, opened 2023-02-21
E2E tests: run bot and backend as docker containers

rzadp:
Currently, in E2E tests the matrix bot and backend start as normal local processes.
This PR starts them as docker containers.

I added a docker network for tests so that the bot&faucet containers can communicate with the pokadot&matrix containers.

### Why?

Because currently the docker containers are not tested (only that they successfully build).
But in production we run the docker containers so this is what we should be testing.


Bullrich (approved):
Seems reasonable 😄

mutantcornholio (approved):

alvicsam (approved):

PR #210, opened 2023-02-21
Migrated to Captcha v2

### LINKED ISSUE #209: Implement Captcha v2
After discussing with @rzadp we decided to use Captcha V2. It's more straightforward and easy to use.

Bullrich:
After discussing with @rzadp we decided that Captcha V2 is more easy to use as V3 requires more configuration server side.

I have migrated to Captcha V2. It works on mobile and normal size screens, and also sets up dark mode when activated.

This commit closes #209


Bullrich:
Preview of the new captcha v2

https://user-images.githubusercontent.com/8524599/220370486-7088ead5-5c2d-43b6-8960-3e54da0f2f5b.mp4

client/src/lib/CaptchaV2.svelte
```diff
@@ -0,0 +1,40 @@
+<script lang="ts">
+  import {createEventDispatcher} from "svelte";
+  export let captchaKey: string;
+
+  const dispatch = createEventDispatcher<{ token: string }>();
+
+  const captchaId = 'captcha_element';
+
+  window.captchaLoaded = () => {
```
rzadp, line 9: I have 0 experience with Svelte. Is it normal to add things globally to `window`? Seems to me like something to avoid.

Bullrich: Not really, but the captcha api expect windows methods, so I don't have any other way of exposing a method for it to call.

If I would be able to pass a reference to a method I would, but unfortunately it doesn't work :(

client/src/lib/CaptchaV2.svelte
```diff
@@ -0,0 +1,40 @@
...
+    const darkTheme = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches;
+    const mobileScreen = window.innerHeight > window.innerWidth;
+
+    // @ts-ignore
```
rzadp, line 13: I would suggested to always specify exactly which rule is supposed to be ignored.

rzadp (approved):

client/src/lib/CaptchaV2.svelte
Bullrich: I found a way to evade it by using the window reference. That will work!

Thanks for this input

alvicsam (approved):

mordamax (approved):

Issue #205, opened 2023-02-21
Operate simultaneously as a matrix bot and external API


rzadp:
Currently, a single deployment of the faucet backend operates in one of two modes:

| Mode | Description |
|---|---|
| `EXTERNAL_ACCESS: false` (default) | The backend expects requests to be coming from the (trusted) matrix bot. We trust matrix account registration to be the first line defense against spam. |
| `EXTERNAL_ACCESS: true` | The backend expects requests to be coming from external (untrusted) API calls. We add captcha as a first line defense against spam |

I'm opening this issue to discuss the matter of changing the backend so that it could operate in both modes simultaneously.

### Why it could be beneficial

1. Decrease the amount on deployments

It is probably safer to run the two different modes in separate deployments, in order to separate the risk of draining the faucet.
But if we want to add the API mode to every environment, that we would **double the amount of deployments**.
If we tolerate the risks (e.g. staging?), or add two separate treasuries in code, we could reduce the amount of new deployments needed.

2. Simplify testing

It would be easier to test a single instance in different ways, than to test two different instances.
Can relate to E2E tests, local development, manually testing staging.

3. We're not losing anything.

We would not lose the possibility to operate in these two modes separately.
No matter the code changes, you can always deploy the backend without exposing any external ports (so matrix bot only), or deploy the backend container without the bot container (so external API only).

### Possible ways

1. Detect where the request is coming from

If we can detect automatically whether the request is coming from the sibling docker container or not, we can use it to deduct whether the request is coming from the bot or not.

(I spent some time to explore this direction. As you dig into it, there are more and more caveats and now I think this is NOT the right way).

2. Add a secret password/token

We could add a secret password/token shared between the backend and the bot.
Then, if an API requests has that password, the backend knows it's coming from the bot.

Seems simple to implement but would add operation complexity to manage the secret.

3. Leave it as it is

Also an option - we could not make any changes and leave it as it is if we're happy with current approach.

rzadp:
@paritytech/opstooling does anyone has any thoughts on this?

Bullrich:
I like option 2 as it requires the less involment from a devops side.

Can we have the secret being set in an env variable on build and then simply set that same secret on the bot?

rzadp:
> Can we have the secret being set in an env variable on build and then simply set that same secret on the bot?

I think so, that would be nice because it wouldn't need a runtime variable.

Bullrich:
[...]

PR #211, opened 2023-02-21
Divided endpoints based on faucet or drip bot

Bullrich:
This PR tries to divide the endpoints into two different methods.

It adds a couple of improvements to reduce duplicated code.

# Endpoints

It created the following endpoints:
- `faucet/`
- `drip-v2/`

I kept the default drip endpoint to not break the already existing implementation, but if `drip-v2` works, we could replace `drip` with that one.

The `faucet` endpoint only handles the expected values coming from the faucet clients (!ink docs and the web faucet) while the drip endpoint handles the other part of the requests.

This way we remove ifs from the code and can also isolate the endpoints.

It adds a couple of improvements to reduce duplicated code.

## Middleware

```typescript
const addressMiddleware = (
  req: Request<unknown, DripResponse, Partial<DripRequestType>>,
  res: Response,
  next: NextFunction,
): void => {
  if (!req.body.address) {
    return missingParameterError(res, "address");
  }
  next();
};
```

It creates this middleware to ensure that the field address is not empty.

## Status code in errors

As the current implementation doesn't return an error status code, I add them so the system can show an error on the request. I also added HTTP codes based on the type of errors.

```typescript
res.status(400).send({ error: "Operation failed." });
```

## Missing field error

I simplified the code of a missing variable into a method that is void (so it can be returned too)
```typescript
const missingParameterError = (res: Response<DripErrorResponse>, parameter: string): void => {
  res.status(400).send({ error: `Missing parameter: '${parameter}'` });
};
```

### DIFF
--- a/src/server/routes/actions.ts
+++ b/src/server/routes/actions.ts
@@ -33,7 +33,10 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
     });
 });
 
-const missingParameterError = (res: Response<DripErrorResponse>, parameter: keyof BotDripRequestType | keyof FaucetRequestType): void => {
+const missingParameterError = (
+  res: Response<DripErrorResponse>,
+  parameter: keyof BotDripRequestType | keyof FaucetRequestType,
+): void => {
   res.status(400).send({ error: `Missing parameter: '${parameter}'` });
 };
 
--- a/src/server/routes/actions.ts
+++ b/src/server/routes/actions.ts
@@ -4,7 +4,7 @@ import express, { NextFunction, Request, Response } from "express";
 import { logger } from "../../logger";
 import {
   BalanceResponse,
-  BotDripRequestType,
+  BotRequestType,
   DripErrorResponse,
   DripRequestType,
   DripResponse,
@@ -35,7 +35,7 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
 
 const missingParameterError = (
   res: Response<DripErrorResponse>,
-  parameter: keyof BotDripRequestType | keyof FaucetRequestType,
+  parameter: keyof BotRequestType | keyof FaucetRequestType,
 ): void => {
   res.status(400).send({ error: `Missing parameter: '${parameter}'` });
 };
@@ -51,9 +51,11 @@ const addressMiddleware = (
   next();
 };
 
+type PartialDrip<T extends FaucetRequestType | BotRequestType> = Partial<T> & Pick<T, "address">;
+
 const dripRequestHandler = new DripRequestHandler(actions, storage, recaptchaService);
 
-router.post<unknown, DripResponse, FaucetRequestType>("/drip/web", addressMiddleware, async (req, res) => {
+router.post<unknown, DripResponse, PartialDrip<FaucetRequestType>>("/drip/web", addressMiddleware, async (req, res) => {
   const { address, parachain_id, recaptcha } = req.body;
   if (!recaptcha) {
     return missingParameterError(res, "recaptcha");
@@ -79,7 +81,7 @@ router.post<unknown, DripResponse, FaucetRequestType>("/drip/web", addressMiddle
   }
 });
 
-router.post<unknown, DripResponse, BotDripRequestType>("/drip/bot", addressMiddleware, async (req, res) => {
+router.post<unknown, DripResponse, PartialDrip<BotRequestType>>("/drip/bot", addressMiddleware, async (req, res) => {
   // Do not allow this endpoint to be accessed from outside. Should only be for the bot
   if (config.Get("EXTERNAL_ACCESS")) {
     return res.status(503).send({ error: "Endpoint unavailable" });
--- a/src/server/routes/actions.ts
+++ b/src/server/routes/actions.ts
@@ -33,7 +33,7 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
     });
 });
 
-const missingParameterError = (res: Response<DripErrorResponse>, parameter: string): void => {
+const missingParameterError = (res: Response<DripErrorResponse>, parameter: keyof BotDripRequestType | keyof FaucetRequestType): void => {
   res.status(400).send({ error: `Missing parameter: '${parameter}'` });
 };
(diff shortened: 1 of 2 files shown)


src/server/routes/actions.ts
```diff
@@ -26,19 +33,83 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
     });
 });
 
+const missingParameterError = (res: Response<DripErrorResponse>, parameter: string): void => {
```
rzadp, line 36: Good idea to extract this.
While we're at it, we could enforce stricter types to catch potential mistakes in the future:

```suggestion
const missingParameterError = (res: Response<DripErrorResponse>, parameter: keyof BotDripRequestType | keyof FaucetRequestType): void => {
```

src/server/routes/actions.ts
```diff
@@ -26,19 +33,83 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
...
+): void => {
+  if (!req.body.address) {
+    return missingParameterError(res, "address");
+  }
+  next();
+};
+
 const dripRequestHandler = new DripRequestHandler(actions, storage, recaptchaService);
 
+router.post<unknown, DripResponse, FaucetRequestType>("/faucet", addressMiddleware, async (req, res) => {
```
rzadp, line 53: Consider having a Partial here same as the previous endpoint, because no matter what types we put here, we could be fooled by typescript because this input comes from external input.

```suggestion
router.post<unknown, DripResponse, Partial<FaucetRequestType>>("/faucet", addressMiddleware, async (req, res) => {
```

For example right now you're treating `address` as defined, but it actually could be undefined if the external user does not supply it in the API call.

Bullrich: The reason for which I'm not checking address is because the `addressMiddleware` is doing so.

I'm not using `partial` as it still gonna throw typescript compilation errors if I do so because it assumes I didn't check that.

I'm still checking all the variables, I'm just verifying them in shared code.

src/server/routes/actions.ts
Bullrich: Oh I love that! That's such a good idea!

src/server/routes/actions.ts
```diff
@@ -26,19 +33,86 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
...
+      res.send(dripResult);
+    }
+  } catch (e) {
+    logger.error(e);
+    errorCounter.plusOne("other");
+    res.status(500).send({ error: "Operation failed." });
+  }
+});
+
+router.post<unknown, DripResponse, BotDripRequestType>("/drip-v2", addressMiddleware, async (req, res) => {
```
mordamax, line 82: few comments:
1. we're exposing now the public URL which you can use directly to call API endpoint to drip amount to any address and substitute any random sender (as it's not controlled by matrix now) and no captcha.. isn't it potential way for breach? )) does it make sense to check if the current server is indeed `config.Get("EXTERNAL_ACCESS") === true` ? because otherwise this should go with captcha token

2. as I see we don't validate parachain_id anywhere, so might worth validate user input?

src/server/routes/actions.ts
```diff
@@ -26,19 +33,86 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
...
+): void => {
+  if (!req.body.address) {
+    return missingParameterError(res, "address");
+  }
+  next();
+};
+
 const dripRequestHandler = new DripRequestHandler(actions, storage, recaptchaService);
 
+router.post<unknown, DripResponse, FaucetRequestType>("/faucet", addressMiddleware, async (req, res) => {
```
mordamax, line 56: nit: could we change a name to something which is not duplicating the project name and has some actionable meaning? some kind of `/drip/web` `/drip/bot` or something else which would make it more clear from the url

mordamax (changes requested):

src/server/routes/actions.ts
Bullrich: Oh god! You are right, I didn't thought about that.

I'm gonna add some kind of check to it to ensure that it doesn't go through.

Quoting @mutantcornholio: 
image

Bullrich:
I applied the requested changes. I'm looking forward to have some kind of implementation like suggested in https://github.com/paritytech/substrate-matrix-faucet/issues/205 to stop us from having to do those external config checks as it would be cleaner.

src/server/routes/actions.ts
rzadp: Ahh I missed the `addressMiddleware` part yesterday - it makes sense now.

src/server/routes/actions.ts
```diff
@@ -26,19 +33,90 @@ router.get<unknown, BalanceResponse>("/balance", (_, res) => {
...
+): void => {
+  if (!req.body.address) {
+    return missingParameterError(res, "address");
+  }
+  next();
+};
+
 const dripRequestHandler = new DripRequestHandler(actions, storage, recaptchaService);
 
+router.post<unknown, DripResponse, FaucetRequestType>("/drip/web", addressMiddleware, async (req, res) => {
```
rzadp, line 56: One last nitpicking on types from my side :D

You check the `address` with `addressMiddleware`, but not the rest.
So how about something like:

```suggestion
router.post<unknown, DripResponse, Partial<FaucetRequestType> & Pick<FaucetRequestType, 'address'>>("/drip/web", addressMiddleware, async (req, res) => {
```

Then the TS types will reflect reality - that `address` is defined, and about the rest we're not sure, and TS will make sure we handle all missing parameters.

rzadp (approved):
Looks all sensible and clean to me.

mordamax (approved):
✅ but there's a room for improvement going forward:
- add integration / e2e tests for server overall, including old and new endpoints
- validate parachain_id as it's user input

Bullrich:
> ✅ but there's a room for improvement going forward:
> 
>     * add integration / e2e tests for server overall, including old and new endpoints
> 
>     * validate parachain_id as it's user input

I agree, I'll make two different issues to handle those cases and work on them as they are a big task by itself.

#212 and #213 will handle then.

PR #214, opened 2023-02-22
Added network dropdown to client

### LINKED ISSUE #207: Add dropdown to web faucet
Add a [discussed](https://github.com/paritytech/substrate-matrix-faucet/pull/204#discussion_r1112918469) with @rzadp and @mordamax, let's add a dropdown to select the network that the faucet is depositing tokens.

To be done after #203

Bullrich:
Added a dropdown to select the network. By current implementation only one network is supported.

This commit closes #207 

## How to test

Go to `client/` and after installing run `VITE_CAPTCHA_KEY=6Ldlmp8kAAAAACEOROGslrJ7gUZjmJIm6WBPUVki VITE_DEMO=true yarn run dev`

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -30,3 +30,8 @@ This will show that the application is running on Demo mode and will not contact
 [![GitHub Pages deploy](https://github.com/paritytech/substrate-matrix-faucet/actions/workflows/deploy-site.yml/badge.svg?event=push)](https://github.com/paritytech/substrate-matrix-faucet/actions/workflows/deploy-site.yml)
 
 We have a GitHub action that evaulates and builds the website, deploying it to GitHub Pages.
+
+## Setting a default parachain
+
+If you want to have a parachain id set by default, you can add the get property with the `parachain` query:
+`https://paritytech.github.io/substrate-matrix-faucet/?parachain=1234`
--- a/client/README.md
+++ b/client/README.md
@@ -21,9 +21,9 @@ This will show that the application is running on Demo mode and will not contact
 
 ## Scripts
 
-- `npm run dev`: To deploy a development instance of the project
-- `npm run build`: To build the project in the `dist` directory
-- `npm run check`: To lint the project of unnecessary code
+- `yarn run dev`: To deploy a development instance of the project
+- `yarn run build`: To build the project in the `dist` directory
+- `yarn run check`: To lint the project of unnecessary code
 
 ## Deployment
 
--- a/client/src/lib/Form.svelte
+++ b/client/src/lib/Form.svelte
@@ -52,6 +52,7 @@
     <div class="form-control w-full max-w-xs col-span-2">
       <input disabled={!useParachain} bind:value={network}
              type="number" placeholder={useParachain ? "Parachain id" : "Using Relay chain"}
+             min="1000" max="9999"
              class="input input-bordered input-primary w-full max-w-xs"/>
     </div>
   </div>
--- a/client/src/lib/Form.svelte
+++ b/client/src/lib/Form.svelte
@@ -6,6 +6,7 @@
   import CaptchaV2 from "./CaptchaV2.svelte";
 
   let address: string = '';
+  let network: string = 'rococo';
   let token: string = '';
   let formValid: boolean;
   $: formValid = !!address && !!token;
@@ -45,13 +46,10 @@
       <label class="label" for="network">
         <span class="label-text">Network</span>
       </label>
-      <select class="select select-bordered select-primary" id="network">
-        <option selected>Roccoco</option>
-        <option>Star Wars</option>
-        <option>Harry Potter</option>
-        <option>Lord of the Rings</option>
-        <option>Planet of the Apes</option>
-        <option>Star Trek</option>
+      <select class="select select-bordered select-primary" id="network" bind:value={network}>
+        <option selected value="rococo">Rococo</option>
+        <option value="bifrost">Bifrost</option>
+        <option value="dali">Dali</option>
       </select>
     </div>
   </div>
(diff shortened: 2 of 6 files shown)


Bullrich:
Screenshot of the new field:
image

oleg-plakida (approved):

sergejparity (approved):

client/src/lib/utils/config.ts
```diff
@@ -1,3 +1,5 @@
 export const CAPTCHA_KEY: string = import.meta.env.VITE_CAPTCHA_KEY;
 export const FAUCET_URL = import.meta.env.VITE_FAUCET_URL;
 export const DEMO_MODE: boolean = !!import.meta.env.VITE_DEMO;
+
+export const CHAINS: { name: string, id: string }[] = [{ name: 'Rococo', id: '1002' }];
```
mordamax, line 5: the parachain_id should not be hardcoded with a network, as there are many other parachain_ids within Rococo
the 1002 is actual only in scope of INK docs. 
so here it should be a just number input, optional, as it's used rarely in `Rococo Testnet faucet`
image
Most often people just drip to their account within rococo chain

client/src/lib/Form.svelte
```diff
@@ -21,58 +22,76 @@
...
     {:then result}
-      <div in:fly="{{ y: 30, duration: 500 }}" class="alert alert-success shadow-lg">
+      <div in:fly={{ y: 30, duration: 500 }} class="alert alert-success shadow-lg">
         <div>
-          <Tick/>
+          <Tick />
           <span>
-            Your funds have been sent.<br/>
+            Your funds have been sent.<br />
             <a href={`https://rococo.subscan.io/extrinsic/${result}`} target="_blank" rel="noreferrer">
```
mordamax, line 76: should be dependent on a network, right?

client/src/lib/utils/config.ts
mordamax: If we start using a list of chains here, I guess this also requires to refactor the config for faucet URLS, as it should be a map
VITE_FAUCET_URL == rococo,
but how we gonna include westend ?

mordamax (changes requested):

Bullrich:
@mordamax I updated the system to accept custom chain ids:

https://user-images.githubusercontent.com/8524599/221863456-6e37a94d-e42f-4bb6-95aa-3fdb1d574f58.mp4

Taking into consideration that we don't even have a `rococo` deployment, shall we first deploy an instance for `rococo` and then, when we prepare a second one for `westend`, we update the UI to have those changes?

client/src/lib/Form.svelte
```diff
@@ -21,47 +24,53 @@
...
+  <div class="inputs-container md:grid md:grid-cols-3 md:gap-4 ">
+    <div class="form-control">
+      <label class="label cursor-pointer">
+        <span class="label-text">Use parachain</span>
+        <input type="checkbox" bind:checked={useParachain} class="checkbox checkbox-primary"/>
+      </label>
+    </div>
+    <div class="form-control w-full max-w-xs col-span-2">
+      <input disabled={!useParachain} bind:value={network}
+             type="number" placeholder={useParachain ? "Parachain id" : "Using Relay chain"}
```
mordamax, line 54:
```suggestion
             type="number" placeholder={useParachain ? "Parachain id" : "Using Relay chain"}
             min="1000" max="9999"
```

mordamax:
> @mordamax I updated the system to accept custom chain ids:
> 
> Taking into consideration that we don't even have a `rococo` deployment, shall we first deploy an instance for `rococo` and then, when we prepare a second one for `westend`, we update the UI to have those changes?

depends, right? do we want to deploy each network separatelly, and have configs in ENV? or have only 1 web server for all kind of networks hardcoded?

For the easiest way i would go with 1 server and hardcoded map of networks & backend api endpoints per network, but if we look from the SEO standpoint - may be makes sense to have each network to be individual page with its own Logos, H1ns, description etc... however this could also be done within 1 server (deployment)

mordamax (approved):
\+ but looks like it requires a bit more work prior launching it, since current faucet works only with Rococo, but it doesn't mention this on UI

client/src/lib/Form.svelte
```diff
@@ -21,47 +24,54 @@
...
+  </div>
+  <div class="inputs-container md:grid md:grid-cols-3 md:gap-4 ">
+    <div class="form-control">
+      <label class="label cursor-pointer">
+        <span class="label-text">Use parachain</span>
+        <input type="checkbox" bind:checked={useParachain} class="checkbox checkbox-primary"/>
+      </label>
+    </div>
+    <div class="form-control w-full max-w-xs col-span-2">
+      <input disabled={!useParachain} bind:value={network}
```
mordamax, line 53: i guess next PRs when we implement multinetwork, this will be renamed to `bind:value={parachain_id}`, and network will be a list with { rococo, westend }

client/README.md
```diff
@@ -21,12 +21,17 @@ This will show that the application is running on Demo mode and will not contact
...
 ## Deployment
 
 [![GitHub Pages deploy](https://github.com/paritytech/substrate-matrix-faucet/actions/workflows/deploy-site.yml/badge.svg?event=push)](https://github.com/paritytech/substrate-matrix-faucet/actions/workflows/deploy-site.yml)
 
 We have a GitHub action that evaulates and builds the website, deploying it to GitHub Pages.
+
+## Setting a default parachain
+
+If you want to have a parachain id set by default, you can add the get property with the `parachain` query:
+`https://paritytech.github.io/substrate-matrix-faucet/?parachain=1234`
```
mordamax, line 37: https://paritytech.github.io/substrate-matrix-faucet/rococo?parachain_id=1234
https://paritytech.github.io/substrate-matrix-faucet/westend?parachain_id=1234

Issue #215, opened 2023-02-24
Migrate existing ink! docs faucet to the new `/drip/web` endpoint


rzadp:
So that the existing `/drip` can be removed.

PR #216, opened 2023-02-24
created tests for endpoints

Bullrich:
Created tests to be run on all the available endpoints and options


src/server/routes/actions.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+      mockHandleRequest.mockRejectedValueOnce("random error in /bot");
+      const res = await request(app).post("/drip/bot").send({ address: "example", amount: "0", sender: "sender" });
+      expect(res.status).toBe(500);
+      expect(res.body.error).toBe("Operation failed.");
+      expect(mockLoggerError).toHaveBeenCalledWith("random error in /bot");
+    });
+  });
+});
+
+describe("/drip tests", () => {
```
rzadp, line 170: Since https://github.com/paritytech/substrate-matrix-faucet/issues/215 is done, we can remove those tests and this endpoint.

.github/workflows/javascript-tests.yml
```diff
@@ -0,0 +1,21 @@
...
+        with:
+          node-version: 16
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Run tests
+        run: yarn test
+      - name: Run lint
+        run: yarn lint
+      - name: Build the project
+        run: yarn build
```
alvicsam, line 21: But lint and test are already in gitlab pipeline, why not add build there?

Bullrich: Oh that's true. Removing it then

Bullrich: Or would it make more sense to run the testing in GitHub actions and the deployment in GitLab?

src/server/routes/actions.spec.ts
```diff
@@ -0,0 +1,279 @@
+// we need the mock methods to be before the imports for jest to run
```
rzadp, line 1: Would it be a good idea to move this exception to `.eslintrc.cjs` (or to `opstooling-js-style` eventually)? (If the file matches `*.spec.ts`, then we disable `import/first`).

.github/workflows/javascript-tests.yml
alvicsam: Since the deployment is already in Gitlab I think it makes sense to have all CI jobs there

src/server/routes/actions.spec.ts
Bullrich: I would say no, let's do it in a per basis case, as in most cases we don't really mock imports, or we are able to mock it in a more effective way.

The only reason for which I had to do this is because the imports initialize code instead of having an initializer, in any other case I would be able to simply mock later.

src/server/routes/actions.spec.ts
Bullrich: Let's merge it for now and remove it in the next PR: #217

This way, if, for some reason, we need to revert, we can use the tests.

What do you think?

rzadp: Yes, agree 👍

src/server/routes/actions.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+      expect(res.status).toBe(400);
+    });
+
+    test("should fail with no sender", async () => {
+      const res = await request(app).post("/drip").send({ address: "example", amount: "100" });
+      expect(res.body.error).toBe(parameterError("sender"));
+      expect(res.status).toBe(400);
+    });
+
+    test("should fail with no sender", async () => {
```
rzadp, line 263: Copy pasted name?

Bullrich: Yup, and it seems I also had the same problem in the previous test scope.

rzadp (approved):

PR #219, opened 2023-02-27
[WIP] Merging bot-based and web-api-based faucet deployments

rzadp:
My take at changes required to achieve https://github.com/paritytech/substrate-matrix-faucet/issues/205

Sending a preview to discuss, cannot be merged (at least not before coordinating the change of environment variables)

### Changes

- Changed some occurences of `server` to `backend`. (the names `server` and `backend` are used interchangeably all across the codebase. I think `backend` makes a little more sense so I changed it in some places I touched to make it more consistent.
- Split the backend config spec into 3 parts:
  - `BACKEND` - the common configuration
  - `BOTBACKEND` - the configuration for the backend of the matrix bot
  - `WEBBACKEND` - the configuration for the backend with the external API endpoint
  - I attempted to split this into 3 different `env.*.config.json` files, but the results were not better.
- The backend starts the different modes based on set environment vars:
  - if `BOTBACKEND_FAUCET_ACCOUNT_MNEMONIC` is set, we start the backend for the matrix bot
  - if `WEBBACKEND_FAUCET_ACCOUNT_MNEMONIC` is set, we start the external API endpoint
  - We can start both modes if both are defined
  - Startup fails if none is defined.


src/server/routes/actions.ts
```diff
@@ -53,61 +37,104 @@ const addressMiddleware = (
 
 type PartialDrip<T extends FaucetRequestType | BotRequestType> = Partial<T> & Pick<T, "address">;
 
-const dripRequestHandler = new DripRequestHandler(actions, storage, recaptchaService);
+export const createActionsRouter = (
+  opts: { type: "bot" } | { type: "web"; webConfig: WebBackendConfig },
+  actions: Actions,
+) => {
```
Bullrich, line 43: Correct me if I'm wrong.

This checks if it's either a bot or a web deployment, and based on that it only enables ONE of the endpoints. Right?

rzadp: Yes, that was it.

PR #220, opened 2023-02-28
added number pattern to parachain input

Bullrich:
Added the field `pattern="\d*"` to the network input which, on phones, brings up the numeric keyboard instead of the regular keyboard.

Just a very small QoL improvement.


mordamax (approved):
❤️

PR #221, opened 2023-03-03
Allow both bot and web API in a single deployment

rzadp:
- The matrix bot and the web API can be run in the same container in a single deployment
  - (Currently it's either the bot, or the web API steered by the `EXTERNAL_ACCESS` var)
- Removed the `/drip/bot` endpoint, instead the bot drips directly using functionality re-used with the `/drip/web` endpoint
- There are no changes in any of the environment variables, so no re-configuration of existing deployments should be needed

### Summarizing the code changes

- Merged the `bot_injected.Dockerfile` and `server_injected.Dockerfile` into a single `faucet_injected.Dockerfile`

The dockerfiles were virtually identical anyways, the only meaningful difference being the start command `CMD`, which can be specified on runtime.

- By default, run the bot&api simultaneously

Currently the project relied on separating the bot logic and the dripping logic by an API call to `/drip/bot`.
Having added the `/drip/web` and made the API externally accessible, we have to make sure to protect the API and recognize trusted requests from the bot from untrusted requests from external API calls.

Here we remove the bot endpoint entirely and the bot does it's own dripping.

The ability to run the bot and the API separately still remains - simply by changing the runtime command, as shown in `docker-compose.rococo-external.yml`.

- Merged `env.bot.config.json` and `env.server.config.json` into a single file `env.faucet.config.json` (but preserving the separation between the two)

This was from a practical need, because `confmgr` insists on being a singleton with only one config spec at a time, so that was the only way to have those two specs together in a single process.
But the bot and server configs are validated separately with [a little bit of hacks](https://github.com/paritytech/substrate-matrix-faucet/blob/07a577dc2fcc2d4f53e26bfb98bae1dc80e9285a/src/config.ts#L28-L54), so the ability to run only a bot with invalid/missing server config and vice versa still remains.

- Made sure to use the Polkadot account as a singleton

When using the same Polkadot account in the bot and the web API, we might run into an issue with duplicated nonces. Here we make sure the Polkadot account is a singleton to alleviate this potential problem.


src/start.ts
```diff
@@ -0,0 +1,5 @@
+import { startBot } from "./bot";
+import { startServer } from "./server";
+
+startBot();
+startServer();
```
Bullrich, line 5: If I get this right, the `yarn start` script will start both the bot and the server in the same instance?

And the bot doesn't require the server anymore?

Also, minor question but just for convention, wouldn't `index.ts` be a better name for the entry script?

src/server/routes/actions.spec.ts
```diff
@@ -11,158 +12,115 @@ import request from "supertest";
...
 
 let app: Express;
 
 const parameterError = (parameter: keyof BotRequestType | keyof FaucetRequestType) =>
   `Missing parameter: '${parameter}'`;
 
-describe("/drip/* tests", () => {
+describe("/drip/web tests", () => {
   beforeEach(() => {
-    jest.resetAllMocks();
```
Bullrich, line 54: It is still working without this? 😮

src/server/routes/actions.spec.ts
```diff
@@ -11,158 +12,115 @@ import request from "supertest";
...
 jest.mock("../../logger", () => {
   return { logger: { error: mockLoggerError } };
 });
 
-jest.mock("../services/DripRequestHandler", () => {
+jest.mock("../../dripper/DripRequestHandler", () => {
   return {
+    getDripRequestHandlerInstance: jest.fn().mockImplementation(() => {
+      return { handleRequest: mockHandleRequest };
+    }),
```
Bullrich, line 25: I didn't wanted to add a factory method, but I will say, I **really was tempted to do so**.

src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -25,12 +25,12 @@ const rpcTimeout = (service: string) => {
...
 };
 
-export class Actions {
+export class PolkadotActions {
   account: KeyringPair | undefined;
   #faucetBalance: number | undefined;
 
   constructor() {
-    logger.info("🤖 Beep bop - Creating the bot's account");
+    logger.info("🚰 Plip plop - Creating the faucets's account");
```
Bullrich, line 33: Omg I love it. Kudos for this.

src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -25,12 +25,12 @@ const rpcTimeout = (service: string) => {
   }, timeout);
 };
 
-export class Actions {
+export class PolkadotActions {
```
Bullrich, line 28: Good idea to rename it. The previous name was a bit too abstract.

Bullrich (commented):
Sorry, I was very hooked with it and I gave some comments before it was ready.

Looks amazing!

rzadp:
@paritytech/opstooling Here is my 2nd take at https://github.com/paritytech/substrate-matrix-faucet/issues/205 superseding https://github.com/paritytech/substrate-matrix-faucet/pull/219.

src/start.ts
rzadp: > If I get this right, the yarn start script will start both the bot and the server in the same instance?

Correct.

> And the bot doesn't require the server anymore?

Yes, but we still probably want to run both - for one reason because the server exposes the prometheus metrics, healthcheck etc.

> Also, minor question but just for convention, wouldn't `index.ts` be a better name for the entry script?

I couldn't use `index.ts` in `{server,bot}/index.ts` because in there I have to have two files - one that starts, one that only exports a function without running it. (one file used to run the single component as en entrypoint, the other file to import the function and run both components)

src/server/routes/actions.spec.ts
mutantcornholio: `resetAllMocks` is actually super confusing =)
`resetAllMocks` and `clearAllMocks` do opposite things from what they are named. 
`resetAllMocks` drops everything including implementation, while `clearAllMocks` only resets the state.

Bullrich: Everyday I learn something new.

Thanks for the lesson!

rzadp: Yeah, that's a good lesson!
I have added `clearAllMocks` there - since I believe that was the original intention.

src/start.ts
rzadp: I guess I could have a single `src/{server,bot}/index.ts` as an entrypoint, and then in `src/index.ts` do:

```
import 'server/index.ts'
import 'bot/index.ts'
```

That would fire those function without the need to import a function and invoke it.

env.faucet.config.json
```diff
@@ -58,6 +58,60 @@
...
+        "description": "token unit for the network",
+        "default": "UNIT",
+        "type": "string"
+      },
+      "FAUCET_IGNORE_LIST": {
+        "description": "A list of Matrix accounts that will be silently (but logged) ignored, comma separated. Example: \"@alice:matrix.org,@bob:domain.com\"",
+        "default": "",
+        "type": "string"
+      },
+      "DEPLOYED_REF": {
```
mordamax, line 105: i think these configs ideally should be merged together under SMF root, as still we're duplicating DEPLOYED_REF and potentially other configs. While the deployment is the same, the configs could be de-duplicated as the source anyway the same

env.faucet.config.json
```diff
@@ -58,6 +58,60 @@
...
+        "required": true,
+        "regexp": "^@.*:.*$",
+        "type": "string"
+      },
+      "MATRIX_SERVER": {
+        "description": "matrix server url",
+        "default": "https://matrix.org",
+        "type": "string"
+      },
+      "NETWORK_DECIMALS": {
```
mordamax, line 90: also dupe

src/server/routes/actions.ts
```diff
@@ -53,9 +50,10 @@ const addressMiddleware = (
 
 type PartialDrip<T extends FaucetRequestType | BotRequestType> = Partial<T> & Pick<T, "address">;
 
-const dripRequestHandler = new DripRequestHandler(actions, storage, recaptchaService);
-
 router.post<unknown, DripResponse, PartialDrip<FaucetRequestType>>("/drip/web", addressMiddleware, async (req, res) => {
+  if (!config.Get("EXTERNAL_ACCESS")) {
```
mordamax, line 54: why do we need this still?

src/start.ts
mordamax: If we follow this Rfc - 
https://github.com/paritytech/opstooling/blob/jp/rfc-modules-organization/rfc-modules-organization.md#module-patterns
then we should avoid using index.ts, however we always can revisit it

env.faucet.config.json
rzadp: 100% agree that this should be de-duplicated.

However I left it duplicated in order to limit changes in a single PR.
Having no changes to env vars, there is less to review and less to think about when deploying it.

I was planning to approach those env vars separately if/when this PR goes through.

src/server/routes/actions.ts
rzadp: Generally I try to have as little as possible behaviour changes when refactoring.

So here I was trying to preserve the current behaviour (`EXTERNAL_ACCESS is not set => endpoint is not working`).

But that was my only reason and I don't insist on it - the change to remove it would be small.

env.faucet.config.json
Bullrich: I'm ok with leaving this change for a second PR, if you have already contemplated these factors

mordamax (approved):
sounds good, if those will be done in next PRs 👍

Bullrich (approved):

rcny (approved):

rzadp:
@rcny Hey, could you double check that all will be in order with the [kubernetes](https://github.com/paritytech/substrate-matrix-faucet/tree/main/kubernetes) setup?

I don't have experience with that. Will the environment vars we have there under `server` and `bot` find their way into the new, now-single docker-compose service?

rcny:
> could you double check that all will be in order with the [kubernetes](https://github.com/paritytech/substrate-matrix-faucet/tree/main/kubernetes?rgh-link-date=2023-03-07T10%3A30%3A24Z) setup?

As I see, [the chart itself](https://github.com/paritytech/helm-charts/tree/main/charts/substrate-faucet) hasn't been updated to reflect the change that the application consists of the single deployment entity now.

So, no, it won't be able to release on Kubernetes with the current chart.

rzadp:
Reviewers, just letting know that I added [a new commit](https://github.com/paritytech/substrate-matrix-faucet/pull/221/commits/000601dd706ed784ef799206f48bafd2fad9a897) changing the gitlab jobs, I missed that earlier - there is no longer a need for a pair of jobs building&publishing the docker image.

cc @PierreBesson

faucet_injected.Dockerfile
```diff
@@ -11,19 +11,19 @@ ARG PROJECT_NAME=""
```
PierreBesson, line 11: I think this file can be renamed from `faucet_injected.Dockerfile` to simply `Dockerfile`.

rzadp: Agree, renamed.

PierreBesson (approved):

Issue #224, opened 2023-03-08
Unable to teleport funds on Westend faucet bot


bakhtin:
I'm getting an error when trying to teleport funds to an account on Westend Collectives parachain. I'm using [Westend faucet bot](https://matrix.to/#/#westend-devops:parity.io).
```
!drip 15zAuboGaAom3e4sA4dRgKokDbkcwSdGkSMjnaptrWX6RY7Y:1001
```
results in error:
```
[2023-03-08T13:08:26.069] [INFO] default - 💸 teleporting tokens
[2023-03-08T13:08:26.079] [ERROR] default - ⭕ An error occured when sending tokens Error: createType(XcmVersionedMultiLocation):: Cannot map Enum JSON, unable to find 'V1' in v2, v3
    at createTypeUnsafe (/backend/node_modules/@polkadot/types-create/cjs/create/type.js:73:18)
    at TypeRegistry.createTypeUnsafe (/backend/node_modules/@polkadot/types/cjs/create/registry.js:323:46)
    at TypeRegistry.createType (/backend/node_modules/@polkadot/types/cjs/create/registry.js:315:17)
    at ApiPromise.createType (/backend/node_modules/@polkadot/api/cjs/base/Decorate.js:122:27)
    at Actions.teleportTokens (/backend/build/src/server/services/Actions.js:103:66)
    at Actions.sendTokens (/backend/build/src/server/services/Actions.js:165:37)
    at /backend/build/src/server/routes/actions.js:45:62
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

PierreBesson:
It looks like this is also affecting Rococo now so this is high priority.

PierreBesson:
See also the same issue on polkadot.js apps: https://github.com/polkadot-js/apps/issues/9127

PierreBesson:
Found on the Rococo faucet room:

query:
```
!drip 5Gc1RWvH2MtDYB7NKrMEuMsVTMApG3rCjYjPpYvZ8yh2yD3X:1002
```

response:
```
createType(XcmVersionedMultiLocation):: Cannot map Enum JSON, unable to find 'V1' in v2, v3
````

Bullrich:
Thanks for the input. If you have any more information that could be helpful please post it here.

@rzadp has created #225 to reproduce the issue

PierreBesson:
[...]

PR #225, opened 2023-03-09
Change teleport mechanism from V1 to V2

rzadp:
- Progresses https://github.com/paritytech/substrate-matrix-faucet/issues/224
- Changed E2E matrix server name to `parity.io` so that in E2E tests we can drip multiple times by being privileged.


Bullrich (approved):
As discussed before, from my side it looks good but I don't know enough to provide valuable feedback

Issue #231, opened 2023-03-14
E2E test for parachain teleportation


rzadp:
The current test for teleporting assets is not quite real:

https://github.com/paritytech/substrate-matrix-faucet/blob/86a71d061a2aede66672285d0a6df7ff2fa7394f/src/faucet.e2e.ts#L73-L81

The bot simply sends a XCM message, and doesn't verify that it actually has the final effect.
Similarly this current test just checks that the bot accepted the message and responded, but doesn't verify the balances before/after. (And it can't do it, because in current E2E setup there is no parachain to teleport to.

I'm creating this issue for discussion and help around this topic.

rzadp:
I've been experimenting with crafting this test and I am quite stuck.

### What I'm doing

I have built locally:
- `polkadot` version `v0.9.39` with `cargo build --profile testnet`
- In same repo, built a parachain with `cargo build --profile testnet -p test-parachain-adder-collator`
- Added resulting binaries `polkadot` and `adder-collator` to PATH
- Run zombienet using native binaries with `./zombienet-macos --provider native spawn examples/0001-small-network.toml` and the following config:

```
# modified zombienet/examples/0001-small-network.toml
[relaychain]
default_image = "docker.io/parity/polkadot:latest"
default_command = "polkadot"
default_args = [ "-lparachain=debug" ]

chain = "rococo-local"

  [[relaychain.nodes]]
  name = "alice"
  validator = true

  [[relaychain.nodes]]
  name = "bob"
  validator = true

[[parachains]]
id = 100

  [parachains.collator]
  name = "collator01"
  command = "adder-collator"
  args = ["-lparachain=debug"]
```

It successfully spins up a local testnet with a parachain:

<img width="1741" alt="Screenshot 2023-03-14 at 12 33 09" src="https://user-images.githubusercontent.com/12039224/225004422-aa97d6d0-42e5-4b3b-ae7c-82bdfea232f7.png">

Next, I'm trying to teleport tokens to parachain 100 in the same way the bot would do it, in the UI (before I move on to automating it in a test).

![Screen Shot 2023-03-14 at 13 53 33-fullpage](https://user-images.githubusercontent.com/12039224/225007461-baadbae8-d1f3-40ec-a111-f2c01aba2e2a.png)

#### Results

The XCM seem to be sent and execution is attempted, but it has no further effect. The balance of Alice does not decrease appropriately and no other effect is observed.

[...]

Issue #212, opened 2023-02-22
Add E2E tests to all endpoints


Bullrich:
This includes:
- `/drip`
- `/drip/bot`
- `/drip/web`

Bullrich:
Hey @mordamax & @rzadp I just realized I don't really know if this can be achieved.

We need a valid captcha which can not be obtained without some kind of UI automation.

There is some tools to do [automated testing](https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do) but I think that's intended for UI testing, not API testing.

rzadp:
In https://github.com/paritytech/substrate-matrix-faucet/pull/219 I have an E2E test which calls the `drip/web` endpoint, and the backend uses a testing captcha secret which makes any captcha token valid.

Maybe it will be enough for an E2E smoke test.

Bullrich:
> In #219 I have an E2E test which calls the `drip/web` endpoint, and the backend uses a testing captcha secret which makes any captcha token valid.
> 
> Maybe it will be enough for an E2E smoke test.

Oh amazing. Will use that knowledge then. Thank you!

mordamax:
the api also has programmatic way to get a token, so may be that could be alternative option with "special" testing keys
https://web.archive.org/web/20230127150134/https://developers.google.com/recaptcha/docs/invisible

```js
grecaptcha.execute('key', {action: 'homepage'}).then((token) => {
               ...
});
```

PR #236, opened 2023-03-17
E2E: Add a parachain, and test asset teleportation

rzadp:
- Change current way of running polkadot to zombienet, in order to add a parachain
- Properly test asset teleportation
- Change the tests so they no longer assert balance of `0` at the beginning, so they no longer fail when re-run
- Add some missing tests, such as checking balance via web endpoint
- Closes https://github.com/paritytech/substrate-matrix-faucet/issues/231
- Closes https://github.com/paritytech/substrate-matrix-faucet/issues/212
- I've run this on CI multiple times to check flakiness. To me it seems non-flaky enough for this kind of test, but it's not at a 100% success rate.

### DIFF
--- a/src/faucet.e2e.ts
+++ b/src/faucet.e2e.ts
@@ -72,7 +72,7 @@ describe("Faucet E2E", () => {
     expect(botMessage.body).toMatch(/^The faucet has (999.*|100.*) UNITs remaining.$/);
   });
 
-  test("The bots drips to a given address", async () => {
+  test("The bot drips to a given address", async () => {
     const userAddress = randomAddress();
     const initialBalance = await getUserBalance(userAddress);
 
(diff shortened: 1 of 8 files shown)


alvicsam (approved):

src/faucet.e2e.ts
```diff
@@ -55,39 +69,85 @@ describe("Faucet E2E", () => {
...
+    );
+  });
+
+  test("The web endpoint teleports to a given address", async () => {
+    const userAddress = randomAddress();
+    const initialBalance = await getUserBalance(userAddress, parachainApi);
+
+    const result = await webEndpoint.post("/drip/web", {
+      address: userAddress,
+      recaptcha: "anything goes",
```
Bullrich, line 140: This is using the public testing captcha, right?

Bullrich (approved):
As always, you are doing an amazing job!

src/faucet.e2e.ts
rzadp: Correct, it's configured [here](https://github.com/paritytech/substrate-matrix-faucet/blob/3178c07e1788f86c3c46dcd58aafacfdc0f1502b/e2e/bootstrap.sh#L70).

mordamax (approved):

PR #238, opened 2023-03-20
client: Added a Dockerimage

Bullrich:
Added a docker image for the client, in the case that we want to deploy the site to any other platform aside from GitHub Pages


rzadp (approved):

client/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+WORKDIR /app
+
+COPY package.json yarn.lock ./
+RUN yarn install --frozen-lockfile
+COPY . ./
+
+ARG DEMO
+ENV VITE_DEMO ${DEMO}
+ARG CAPTCHA_KEY
```
PierreBesson, line 11: Those variables should not be set in docker build args but by docker environment variables. If those environment specific variables are set in the docker image itself, the image will not longer be portable to multiple environments without rebuilding.

PierreBesson (changes requested):
There should be a CI job that pushes this image to `paritytech/faucet-ui` with the same tags as the backend.

PierreBesson (dismissed):

Bullrich:
Hi @PierreBesson! I added the logic to build the image and publish it in `gitlab-ci`.

Could you please review that file and let me know if it is as intended? I'm afraid my knowledge of gitlab's deployment script is not very broad

rzadp:
Should we update `publish-docker-image-description` as well?

mordamax (approved):

sergejparity:
> Should we update `publish-docker-image-description` as well?

Yes, this make sense. I'll add the commit for that.

sergejparity (approved):

PR #239, opened 2023-03-20
client: migrated to SvelteKit

Bullrich:
Migrate the site to run in SvelteKit instead of plain Svelte. This will allow us to run as a Docker image required by #238

Almost all of the code remains the same, the biggest difference is the new location of files (as SvelteKit requires particular directories) or the modification of config files (that were replaced by the new configs).

Another small change is that public variables now need the `PUBLIC_` prefix instead of the `VITE_` prefix, so we also had to change that value in the CI (so, for @paritytech/CI, this is a very small review in that regard).


rcny (approved):

client/README.md
```diff
@@ -11,20 +11,24 @@ Two current options are to [access Matrix and contact a bot](https://wiki.polkad
 ## Development
 
 To develop you need two env variables:
-- `VITE_CAPTCHA_KEY`: The [reCaptcha v3 site key](https://www.google.com/u/1/recaptcha/admin).
-- `VITE_FAUCET_URL`: The endpoint to contact the faucet.
+- `PUBLIC_CAPTCHA_KEY`: The [reCaptcha v2 site key](https://www.google.com/u/0/recaptcha/admin).
+- `PUBLIC_FAUCET_URL`: The endpoint to contact the faucet.
 
-The reason for which these variables have `VITE` as a prefix is a security meassure to not upload any unnecesary data. [More info here](https://vitejs.dev/guide/env-and-mode.html#env-files)
+The reason for which these variables have `PUBLIC_` as a prefix is a security meassure to not upload any unnecesary data. [More info here](PUBLIC)
```
rzadp, line 17: `[More info here](PUBLIC)`

Forgot to fill in the url here?

client/package-lock.json
rzadp, line 1: This introduces both `yarn.lock` and `package-lock`?

client/README.md
Bullrich: Oh right, I thought I pasted it, good catch.

client/package-lock.json
Bullrich: I'll remove the package-lock. Got in by accident

PR #240, opened 2023-03-21
client: added linting & test

Bullrich:
Added linting to the project (it actually already had it, I just didn't use it as it would change to many files).

Ran `yarn run format` which runs the included eslint and prettier (which has pluggins formated for svelte language so I don't know if we can actually use `opstooling-js-style` for this case)

Added CI step for running this new step.


rzadp (approved):

Bullrich:
@rzadp could you please review it again? I have added tests also to this PR

altaua (approved):

PR #243, opened 2023-03-22
fixed captcha race condition on load

### LINKED ISSUE #242: Race condition (?) when loading Recaptcha in the web faucet
Seems like there is a race condition when loading Recaptcha.

It appears when opening the faucet in Firefox.

To reproduce:

1. Start the app

```
cd client
export PUBLIC_FAUCET_URL="https://example.com/test"      
export PUBLIC_CAPTCHA_KEY="6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI"
npm run build && npm run preview 
```

2. Open the site in Firefox

3. Reload the page without cache a couple of times.

Sometimes, the captcha does not load.
Sometimes without any error, sometimes with `TypeError: window.grecaptcha.render is not a function`

cc @Bullrich

Bullrich:
This closes #242

There was a race condition where sometimes the script of the captcha component hasn't loaded on time and it was causing a runtime error.

This commit modifies the code so that all the window methods are set first, and then the script tag is imported, this way we know that the required environment for the captcha is correctly set up.

### DIFF
--- a/client/src/lib/components/CaptchaV2.svelte
+++ b/client/src/lib/components/CaptchaV2.svelte
@@ -1,10 +1,12 @@
 <script lang="ts">
 	import { createEventDispatcher, onMount } from "svelte";
+	import Cross from "./icons/Cross.svelte";
 	export let captchaKey: string;
 
 	const dispatch = createEventDispatcher<{ token: string }>();
 
 	const captchaId = "captcha_element";
+	let captchaError: boolean = false;
 
 	let componentMounted: boolean;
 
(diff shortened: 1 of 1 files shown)


client/src/lib/components/CaptchaV2.svelte
```diff
@@ -6,20 +6,25 @@
...
-		if (window.grecaptcha) {
+	onMount(() => {
+		window.captchaLoaded = () => {
+			const darkTheme =
+				window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").matches;
+			const mobileScreen = window.innerHeight > window.innerWidth;
+
+			if (!window.grecaptcha) {
+				throw new Error("grecaptcha is undefined!");
+			}
```
Bullrich, line 19: Should we force the browser to reload the page if this happen?

mutantcornholio: I wouldn't go down that road, to avoid infinite reloads.

mutantcornholio (approved):

PR #250, opened 2023-03-30
added parachain selection modal

### LINKED ISSUE #247: Input field breaks if a letter is added
To break the input field → enter letters, afterwards no more chars (even numbers) can be input. Only option is to reload the form.

Bullrich:
Added a modal where the user can select a default chain, a predefined chain or add their own.

image

This closes #247

### DIFF
--- a/.github/workflows/tag-client-pr.yml
+++ b/.github/workflows/tag-client-pr.yml
@@ -12,8 +12,10 @@ jobs:
   label-prs:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v3
       - name: Label PR
         run: gh pr edit $PR_NUMBER --add-label $LABEL_NAME
         env:
           PR_NUMBER: ${{ github.event.number }}
           LABEL_NAME: "client"
+          GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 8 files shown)


client/src/lib/utils/networkData.ts
```diff
@@ -0,0 +1,27 @@
...
+	getChainName(id: number): string | null;
+}
+
+export const Rococo: NetworkData = {
+	networkName: "Rococo",
+	chains: [
+		{ name: "Relay Chain", id: -1 },
+		{ name: "One", id: 1 },
+		{ name: "Two", id: 2 },
+		{ name: "Three", id: 3000 }
```
Bullrich, line 18: @paritytech/opstooling which chains should we add here?

rzadp: I don't know how to get a list of all possible ones, but perhaps scrolling through the faucet matrix chats and seeing which parachains people use would be a good trail?

.github/workflows/tag-client-pr.yml
Bullrich, line 1: @paritytech/ci I just added the env var because it was missing for the execution

Bullrich (commented):
PR has been finished. It is ready for review.

For the network value, the relay chain uses `-1` but when doing the request, it checks if it's `< 0`, if so, it just doesn't send the custom chain id.

rzadp (approved):

PR #255, opened 2023-04-04
added parachain id validation on server

### LINKED ISSUE #213: Validate parachain ID
We are not validating the ID of a parachain, just sending to that chain when we get a user request.

We need to verify if such id is correct, maybe by having a list or a configuration?

Bullrich:
Simple check to verify that the parachain id is a number and it is between 1000 and 9999.

Closes #213

### DIFF
--- a/src/faucet.e2e.ts
+++ b/src/faucet.e2e.ts
@@ -3,7 +3,7 @@ import { createTestKeyring } from "@polkadot/keyring";
 import { HttpProvider } from "@polkadot/rpc-provider";
 import { BN } from "@polkadot/util";
 import { randomAsU8a } from "@polkadot/util-crypto";
-import axios, { AxiosError } from "axios";
+import axios from "axios";
 import { until } from "opstooling-js";
 
 const randomAddress = () => createTestKeyring().addFromSeed(randomAsU8a(32)).address;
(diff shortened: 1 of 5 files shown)


src/faucet.e2e.ts
```diff
@@ -150,4 +160,24 @@ describe("Faucet E2E", () => {
...
+    } catch (error) {
+      if (axios.isAxiosError(error)) {
+        const axiosError = error as AxiosError<{ error: string }>;
+        expect(axiosError.response?.data.error).toEqual(
+          "Parachain invalid. Be sure to set a value between 1000 and 9999",
+        );
+      } else {
+        throw error;
+      }
+    }
```
mutantcornholio, line 181: Wouldn't it work somewhat like this?

```ts
const promise = webEndpoint.post("/drip/web", { address: userAddress, recaptcha: "anything goes", parachain_id: "100" });

await expect(promise).rejects.toThrow(AxiosError);
await expect(promise).rejects.toMatchObject({response: {data: {
  error: "Parachain invalid. Be sure to set a value between 1000 and 9999"
}}});
```

Bullrich: It would! Thank you for that!

mordamax: `throw new Error("Unexpected success");` lmao

mordamax (approved):

src/faucet.e2e.ts
```diff
@@ -11,7 +11,7 @@ const randomAddress = () => createTestKeyring().addFromSeed(randomAsU8a(32)).add
 describe("Faucet E2E", () => {
   const matrix = axios.create({ baseURL: "http://localhost:8008" });
   const webEndpoint = axios.create({ baseURL: "http://localhost:5555" });
-  const PARACHAIN_ID = 100; // From the zombienet config.
+  const PARACHAIN_ID = 1000; // From the zombienet config.
```
rzadp, line 14: So we have to change the zombienet config as well - it's [here](https://github.com/paritytech/polkadot-testnet-faucet/blob/570eb9e731bb32f9cb60e0dabf8595d3118ec57a/e2e/zombienet.native.toml#L17).

Bullrich: Is this config from zombie net valid, or was it set only for the e2e tests?

Bullrich: IT PASSED!

mutantcornholio (approved):

PR #260, opened 2023-04-20
Created issue template for faucet problems

Bullrich:
As matrix and web users could have problems with the faucet in the future, I recommend that we have a template to facilitate them when writing issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/faucet.yml
+++ b/.github/ISSUE_TEMPLATE/faucet.yml
@@ -24,7 +24,7 @@ body:
     id: address
     attributes:
       label: Address
-      description: What was the address where you encountered the problem?
+      description: (Optional) What was the address where you encountered the problem?
       placeholder: ex. 5CPkfj09f4WWF4320jjFcFdmCF1CSLbhDeg3gbA
   - type: input
     id: network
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/faucet.yml
```diff
@@ -0,0 +1,42 @@
...
+      label: "Problem"
+      description: Please provide a clear and concise description of the problem.
+    validations:
+      required: true
+  - type: input
+    id: address
+    attributes:
+      label: Address
+      description: What was the address where you encountered the problem?
+      placeholder: ex. 5CPkfj09f4WWF4320jjFcFdmCF1CSLbhDeg3gbA
```
Bullrich, line 28: I added the address in the case that we need to search if the problem is related to the address itself (has enough tokens or is blacklisted).

This won't be a security issue, right?

rzadp: My only suggestion is to explicitly mark it as optional.
It's not configured to be required already, but I would make it more apparent, e.g.:

`(optional) What was the address where you encountered the problem?`

rzadp (approved):

sergejparity (approved):

PR #269, opened 2023-05-23
Card redesign

### LINKED ISSUE #268: Create new form card

Bullrich:
This closes #268

Created new UI for the card which brings the new design. 

This design is ONLY for the card, navbar and background will be done in different tasks.


rzadp:
The UI looks great!
I have some comments about the UX:

First, should the `Use custom chain id` and `Use preselected chains` labels be reversed?

Screenshot 2023-05-24 at 10 23 57

Here I have a button saying `Use custom chain id` - but I am already using custom chain id.

---

Second - Type a custom chain id, then switch to preselected chains. I have an ugly `null` there, and the submit button is enabled.

Screenshot 2023-05-24 at 10 27 16

It should probably select the first one from the list, or at least disable the submit button.

---

Third - about the selector.

Screenshot 2023-05-24 at 10 29 11

- Selecting (that is clicking on one of the chains) does not close the dropdown. I would be expecting the dropdown to be closed.
- Clicking on the selector does not close the dropdown as well, contrary to my expectations.
Screenshot 2023-05-24 at 10 30 36

Bullrich:
Thanks for your review @rzadp.

I have fixed all the issues you have mentioned

rzadp (approved):

PR #290, opened 2023-06-09
question btn hidden on mobile

Bullrich:
Made the "Questions" button hide in mobile. This is because too many buttons in the UI sometimes overflow the width on the screen.

In my phone which has a small screen I could see the screen overflowing.

With the btn:

Without the btn:

### DIFF
--- a/client/src/lib/components/NavBar.svelte
+++ b/client/src/lib/components/NavBar.svelte
@@ -33,7 +33,7 @@
 
 <style lang="postcss">
 	.questions-btn {
-		@apply btn text-white hover:text-opacity-70;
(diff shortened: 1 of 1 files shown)


mutantcornholio:
Is it useless on mobile? Should we consider moving it somewhere instead of hiding?

Bullrich:
> Is it useless on mobile? Should we consider moving it somewhere instead of hiding?

It just scrolls down to the FAQ for you. It doesn't really redirect you somewhere.

On mobile you will still be able to manually scroll down and go to the faq.

mutantcornholio (approved):
In that case, sounds reasonable

PR #300, opened 2023-06-14
Apply common codestyle to client

mutantcornholio:
This is a step to move client/ to src/, furthermore joining what was three different repos before.
I tried to minimize the amount of changes here, so there is still `client/.eslintrc.cjs`, and still a separate `package.json`.


mutantcornholio:
finally!

Screen Shot 2023-06-15 at 12 22 18
Screen Shot 2023-06-15 at 12 22 25

sergejparity (approved):

.github/workflows/deploy-site.yml
```diff
@@ -25,7 +25,6 @@ jobs:
           node-version: 18
       - run: yarn install --frozen-lockfile
       - run: yarn run check
-      - run: yarn run lint
```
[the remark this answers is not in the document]
mutantcornholio: Both server and client are being checked together now: https://github.com/paritytech/polkadot-testnet-faucet/pull/300/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R18-R21

src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -189,7 +189,7 @@ export class PolkadotActions {
     if (dripTimeout) clearTimeout(dripTimeout);
 
     if (isDripSuccessResponse(result)) {
-      await this.updateFaucetBalance().then(() => logger.info("Refreshed the faucet balance 💰"));
+      await this.updateFaucetBalance().then(() => logger.info("Refresheªd the faucet balance 💰"));
```
rzadp, line 192: 🧐

mutantcornholio: Whoa, thanks!

rzadp (approved):

client/src/lib/utils/faucetRequest.ts
```diff
@@ -1,57 +1,54 @@
...
+  network: NetworkData,
+  parachain?: number,
 ): Promise<string> {
-	if (DEMO) {
-		return boilerplateRequest(address, recaptcha);
-	}
-	const chain = parachain && parachain > 0 ? parachain.toString() : undefined;
-	return faucetRequest(address, recaptcha, network, chain);
+  if (DEMO !== undefined && DEMO !== "") {
+    return await boilerplateRequest(address);
```
Bullrich, line 12: Isn't `if (DEMO)` achieving the same result (and also including if DEMO is not null?

client/src/lib/utils/faucetRequest.ts
```diff
@@ -1,57 +1,54 @@
...
+  }
+  const chain = parachain && parachain > 0 ? parachain.toString() : undefined;
+  return await faucetRequest(address, recaptcha, network, chain);
```
Bullrich, line 15: Do we get any benefit by returning the await of a promise vs returning the promise itself?

client/src/lib/utils/faucetRequest.ts
```diff
@@ -1,57 +1,54 @@
...
+    method: "POST",
+    body: JSON.stringify(body),
+    headers: { Accept: "application/json", "Content-Type": "application/json" },
+  });
+  // FIXME
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const result = await fetchResult.json();
+  if ("error" in result) {
+    // FIXME
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
Bullrich, line 40: We should generate an issue from this to fix it.

client/src/lib/utils/networkData.ts
```diff
@@ -2,60 +2,53 @@ import { base } from "$app/paths";
...
+  { network: Westend, url: `${base as string}/westend` },
 ];
+
+export function getChainName(network: NetworkData, id: number): string | null {
+  const index = network.chains.findIndex((ch) => ch.id === id);
+  if (index < 0) {
+    return null;
+  }
+  return network.chains[index].name;
+}
```
Bullrich, line 54: Good abstraction

client/tsconfig.json
```diff
@@ -1,17 +1,28 @@
...
+  "include": [
+    ".svelte-kit/ambient.d.ts",
+    "./src/**/*.js",
+    "./src/**/*.ts",
+    "./src/**/*.svelte",
+    "./tests/**/*.js",
+    "./tests/**/*.ts",
+    "./tests/**/*.svelte",
+    "playwright.config.ts",
+    "vite.config.ts"
```
Bullrich, line 22: Should these files be included?

Playwright has it's own tsconfig (internally) and I don't want to bloat the site binary with extra information

client/src/lib/utils/faucetRequest.ts
mutantcornholio: > Isn't `if (DEMO)` achieving the same result 

Yes, but implicit boolean casts violate our codestyle

> (and also including if DEMO is not null?

how an env variable would end up being null?

client/src/lib/utils/faucetRequest.ts
mutantcornholio: We get +1 line in stacktraces, which might shed light on the error origin in some cases.  
I wouldn't call myself a big fan of the rule, but I would like to keep it rather than to change it.

client/tsconfig.json
mutantcornholio: > and I don't want to bloat the site binary with extra information

Wait what? It's tsconfig, it doesn't dictate builder what to bundle, it just says which files should be checked.

mordamax (approved):

Bullrich (approved):
All my doubts have been cleared. Green light from my side

PR #309, opened 2023-06-20
Removing http requests between bot and server

mutantcornholio:
These parts are in the same process now, so the data can be queried
directly

### DIFF
--- a/src/bot/index.ts
+++ b/src/bot/index.ts
@@ -1,5 +1,4 @@
 import { decodeAddress } from "@polkadot/keyring";
-import axios from "axios";
 import dotenv from "dotenv";
 import * as mSDK from "matrix-js-sdk";
 import request from "request";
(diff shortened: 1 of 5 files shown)


src/bot/index.ts
```diff
@@ -106,36 +101,19 @@ bot.on("Room.timeline", (event: mSDK.MatrixEvent) => {
...
+      const balance = await polkadotActions.getBalance();
+
+      sendMessage(
+        roomId,
+        `The faucet has ${Number(balance) / 10 ** networkData.decimals} ${networkData.currency}s remaining.`,
+      );
+    })().catch((e) => {
+      sendMessage(roomId, "An error occurred, please check the server logs.");
+      logger.error("⭕ An error occurred when checking the balance", e);
+    });
```
Bullrich, line 116: Could this maybe be replaced for a try/catch block?

As it is no longer calling a function, it could be cleaner to read maybe?

Bullrich (approved):
Lots of great changes Yuri 💪

src/bot/index.ts
mutantcornholio: The surrounding function (the callback for `MatrixClient#on`) is synchronous

PR #317, opened 2023-07-10
Use testcontainers for e2e tests

### LINKED ISSUE #202: E2E tests: use `testcontainers`?
Following [this suggestion](https://github.com/paritytech/substrate-matrix-faucet/pull/198#issuecomment-1429867889), I spend some time to try and use `testcontainers-node` for the E2E tests instead of the current shell script setup.

However, I run into several issues with it:

- I could not use a docker-compose with supplied environment variables. I suspect the fault of [this bug](https://github.com/testcontainers/testcontainers-node/issues/459) uncovered just 2 days ago.
- Somehow, the docker build cache of the bot&server docker images is not reused between test runs. So, every run of the tests, we're waiting for their `yarn install` steps.
- The container logs are not separated. Every container just logs into `testcontainers:containers` DEBUG scope. So it is not possible to easily dig into the logs of a single container.
- The setup and development does not seem to be getting any simpler.

I think that `testcontainers` might be useful but not in this particular repo. Or maybe I approached it incorrectly.

I have pushed my WIP changes to the [rzadp/testcontainers](https://github.com/paritytech/substrate-matrix-faucet/tree/rzadp/testcontainers) branch, if anyone would like to take a look.

I'm creating this issue as a discussion point.

mutantcornholio:
Key points here:
* application container is built manually, in order to reuse cache
* changed startup, so web part would be last to start, and listening to web port might be considered as "ready" event. This change will be essential for proper service replication. Things that are now being waited for are matrix sync and `isReady` on polkadotjs api
* had to create additional Dockerfile for synapse, as I couldn't solve permissions issue through testcontainers. However, building that is super fast, fine to do every time

Fixes #202


e2e/matrix_container.Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM matrixdotorg/synapse:v1.87.0
+
+# Tried to solve this with testingcontainers, failed
```
rzadp, line 3:
```suggestion
# Tried to solve this with testcontainers, failed
```

src/bot/index.ts
```diff
@@ -183,9 +183,29 @@ bot.on(mSDK.RoomEvent.Timeline, (event: mSDK.MatrixEvent) => {
   }
 });
 
-export const startBot = () => {
-  bot.startClient({ initialSyncLimit: 0 }).catch((e) => logger.error(e));
-};
+export async function startBot(): Promise<void> {
+  // Resolving on error makes allows web server to start even if matrix is imparied
```
rzadp, line 187:
```suggestion
  // Resolving on error allows web server to start even if matrix is imparied
```

src/test/matrixHelpers.ts
```diff
@@ -0,0 +1,103 @@
+import { validatedFetch } from "@eng-automation/js";
+import Joi from "joi";
+
+
+const accessTokenResponseSchema = Joi.object({ acces_token: Joi.string() });
```
rzadp, line 5:
```suggestion
const accessTokenResponseSchema = Joi.object<accessTokenResponse>({ access_token: Joi.string() });
```

rzadp (approved):
Great work!

src/test/matrixHelpers.ts
mutantcornholio: Does it bring anything aside from more characters?

rzadp: You mean the `<accessTokenResponse>` part? It prevents the `acces_token` typo from happening, right now it fails silently.

mutantcornholio: Armed with a "wait, how on earth did it work???" question, I realised that I forgot to put `.required()` to schemas...

mutantcornholio: nice catch

PR #321, opened 2023-07-18
Using @eng-automation/js logger and metrics

mutantcornholio:
* Using our logger instead of log4js
* Using prom-client instead of constructing metric responses by hand
* Added default prom-client node.js metrics. These are mainly focused on
performance, and can shed light on many related problems
* Capturing metrics on log entries


package.json
```diff
@@ -64,13 +64,12 @@
...
     "dotenv": "^16.0.0",
     "express": "4.17.3",
-    "log4js": "6.4.5",
     "matrix-js-sdk": "^26.1.0",
     "request": "^2.88.2",
     "sqlite3": "^5.1.2"
   },
   "devDependencies": {
-    "@eng-automation/js": "^0.0.22",
+    "@eng-automation/js": "^0.0.23",
```
rzadp, line 72: While we're at it - looks like this should be a non-dev dependency.

rzadp (approved):

package.json
mutantcornholio: Definitely!

PR #326, opened 2023-07-21
Trappist faucet + updates

mutantcornholio:
Stopped copying known static urls like PUBLIC_FAUCET_ROCOCO_URL, storing
it in NetworkData now.
Setting PUBLIC_FAUCET_URL to anything non-empty will override all
requests to it.

Trappist doesn't exist on subscan yet, so making it optional.

### DIFF
--- a/client/.env
+++ b/client/.env
@@ -1,6 +1,9 @@
 PUBLIC_DEMO_MODE=
 PUBLIC_CAPTCHA_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
-PUBLIC_FAUCET_ROCOCO_URL=
-PUBLIC_FAUCET_WESTEND_URL=
+
+PUBLIC_FAUCET_URL=
+# uncomment to direct requests to local instance
+# PUBLIC_FAUCET_URL=http://localhost:5555
+
 PUBLIC_ISSUE_LINK=https://github.com/paritytech/polkadot-testnet-faucet/issues/new/choose
 PUBLIC_FORUM="https://forum.polkadot.network/t/experiencing-trouble-accessing-our-rococo-faucet-please-post-here/2952"
--- a/client/src/lib/utils/networkData.ts
+++ b/client/src/lib/utils/networkData.ts
@@ -1,9 +1,17 @@
 import { base } from "$app/paths";
-import { PUBLIC_FAUCET_ROCOCO_URL, PUBLIC_FAUCET_WESTEND_URL } from "$env/static/public";
+import { PUBLIC_FAUCET_URL } from "$env/static/public";
 
 export interface ChainData {
   name: string;
   id: number;
+}
+
+function faucetUrl(defaultUrl: string): string {
+  if (PUBLIC_FAUCET_URL !== "") {
+    return PUBLIC_FAUCET_URL;
+  }
+
+  return defaultUrl;
 }
 
 export interface NetworkData {
@@ -11,12 +19,12 @@ export interface NetworkData {
   currency: string;
   chains: ChainData[];
   endpoint: string;
-  explorer: string;
+  explorer: string | null;
 }
 
 export const Rococo: NetworkData = {
   networkName: "Rococo",
-  currency: "$ROC",
+  currency: "ROC",
   chains: [
     { name: "Rococo Relay Chain", id: -1 },
     { name: "Rockmine", id: 1000 },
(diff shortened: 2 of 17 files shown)


client/tests/rococo.test.ts
```diff
@@ -8,5 +8,5 @@ const chains = [
   { name: "Bridgehub", id: 1013 },
 ];
 
-const test = new FaucetTests("/", "PUBLIC_FAUCET_ROCOCO_URL", "Rococo Faucet", chains);
-test.runTests();
+const tests = new FaucetTests({ faucetName: "Rococo Faucet", chains, url: "/", expectTransactionLink: true });
```
mordamax, line 11: imo would be cool to be explicit there, and instead of root "/" have here "/rococo"
then in the root of app show 3 (or how many) links/cards to supported networks.. so whoever lands there can click needed network

client/.env
```diff
@@ -1,6 +1,9 @@
 PUBLIC_DEMO_MODE=
 PUBLIC_CAPTCHA_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
-PUBLIC_FAUCET_ROCOCO_URL=
-PUBLIC_FAUCET_WESTEND_URL=
+
+PUBLIC_FAUCET_URL=
+# uncomment to direct requests to local instance
+PUBLIC_FAUCET_URL=http://localhost:5555
```
Bullrich, line 6: This is duplicated. Did you intend to do:

```suggestion
PUBLIC_FAUCET_URL=
# uncomment to direct requests to local instance
# PUBLIC_FAUCET_URL=http://localhost:5555
```

src/networkData.ts
```diff
@@ -70,6 +70,17 @@ const versi: NetworkData = {
   rpcEndpoint: "wss://versi-rpc-node-0.parity-versi.parity.io/",
 };
 
+const trappist: NetworkData = {
+  balanceCap: 100,
+  chains: [],
+  currency: "HOP",
+  decimals: 12,
```
mordamax, line 77: do we need to hardcode a currency and decimals? I guess they could be loaded dynamically from the onchain config

Bullrich (approved):

client/src/lib/utils/networkData.ts
```diff
@@ -1,48 +1,65 @@
 import { base } from "$app/paths";
-import { PUBLIC_FAUCET_ROCOCO_URL, PUBLIC_FAUCET_WESTEND_URL } from "$env/static/public";
+import { PUBLIC_FAUCET_URL } from "$env/static/public";
 
 export interface ChainData {
   name: string;
   id: number;
 }
 
+function faucetUrl(defaultUrl: string): string {
```
mordamax, line 9: @mutantcornholio sorry, i don't understand this swap. 
Why can't it leave in PUBLIC_FAUCET_\<network\>_URL and locally .env you make it as localhost:5555 or prod url? 
i think "unsetting" PUBLIC_FAUCET_URL to "" in order to make calls to prod, vs setting prod URL explicitly to PUBLIC_FAUCET_WESTEND_URL would be more clear and expected?

src/networkData.ts
Bullrich: From what I see, [decimals is being used](https://github.com/search?q=repo%3Aparitytech%2Fpolkadot-testnet-faucet%20decimals&type=code) to convert values into human readable or do some big int calculations.

How would you load the amount of decimals from on-chain? 
Taking into consideration that they never change, keeping them hardcoded seems like a practical solution to me

client/tests/rococo.test.ts
mordamax: yes, but they still can point directly to a needed faucet, so there will be same amount of clicks, you just provide /rococo in appropriate docs, but there will be better navigation and clear distinction 

who is "they" btw? )

client/src/lib/utils/networkData.ts
mutantcornholio: My reasoning is, to reduce the amount of necessary variables, with the growth of faucet deployments.   

These URLs are static, and AFAIK the only valid reason to override them, is to route faucet client to dev faucet backend.  
Adding a new production deployment should not alter development.  
Forcing to provide a separate variable for each deployment that we have just does not sound right.

src/networkData.ts
mutantcornholio: I don't feel that loading them dynamically would be an improvement. It would make code considerably more complex, while not giving much in return.  

We still would need these static configs, at least for `explorer`, `rpcEndpoint`, `balanceCap` and `dripAmount` properties. Cutting them in half, while adding all this dynamic initialisation doesn't look like a win to me.

client/tests/rococo.test.ts
mutantcornholio: I agree with Mak, it also bugs me a bit.  
Wouldn't do it here though, as it has to be updated in several places.

Bullrich: I'll write a ticket and we can go to design with a requirement for it

Bullrich: Done: https://github.com/paritytech/polkadot-testnet-faucet/issues/329

alvicsam (approved):

mordamax (approved):

PR #336, opened 2023-08-17
Introducing postgres to faucet

mutantcornholio:
Using TypeORM as probably most popular and mature TS ORM out there

* Removed DripperStorage and its tests, as all logic inside of it became
super trivial.
* Renamed `isValid`/`saveData` functions from DripperStorage to more
explicit `saveDrip`/`hasDrippedToday`.
* Changed DripRequestHandler tests to more unit-tests approach with
mocking dripperStorage.
* Upgrading prettier (that was required to use with decorators properly)
results in insignificant changes accross some files (app.css, app.html,
etc.)

A part of https://github.com/paritytech/polkadot-testnet-faucet/issues/335


mutantcornholio:
> What is your opinion regarding [Prisma](https://www.prisma.io/)

Another option worth considering

package.json
```diff
@@ -62,20 +62,20 @@
     "bs58": "^5.0.0",
     "confmgr": "^1.0.8",
     "cors": "^2.8.5",
-    "crypto": "^1.0.1",
```
rzadp, line 65: Why is this removed? It's used directly in `src/dripper/dripperStorage.ts`

rzadp: Ah ok, I didn't realize this was a `node` thing.

rzadp (approved):

PR #353, opened 2023-10-17
Fix transfer on rococo

mordamax:
transfer is undefined now for some reason (reason -> https://github.com/paritytech/polkadot-sdk/pull/1226)
replaced with trasferKeepAlive

image

The reason is valid, as `transfer` was actually `transfer_allow_death`, while we supposed to use transfer_keep_alive from the start


Bullrich (approved):
Simple lib upgrade 😄

rzadp (approved):

PR #354, opened 2023-10-18
Fix teleport xcm types

mordamax:
Fix for failed teleport in Rococo faucet bot 
image

I couldn't find any release notes regarding this change :(

just went to https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frococo-rpc.polkadot.io#/extrinsics and looked up the xcmPallet.limitedTeleportAssets method, where i saw expected types, which helped to migrate

the only difference is that all networks (westend, polkadot, kusama) except Rococo use similar types but without "Staging" prefix
Weird, but okay... if/when any other network also starts using Staging - it'd be easy to fix by adding them to condition.. still wondering how would I learn about this change in front


rzadp (approved):

mutantcornholio:
This sounds like something that's bound to break once again, if they'll remove the prefix...

PR #371, opened 2024-02-07
add paseo, fix UX, whitelist paseo people separatelly

### LINKED ISSUE #365: Understandability/layout
Hey there,

This is more just general feedback, but it seems to me that "Network" and "Chain" are kind of confusing terms. Instead of mixing relay and parachains I think it would make more sense for the selectors to look like this:

Network     |    Chain
------------------------
Rococo      |    Relay (default), Contracts, Asset Hub, Trappist, etc
Westend    |     Relay (default), Asset Hub, etc
Paseo         |    Relay (default), Asset Hub, Contracts, etc
// etc

Right now it seems a bit confusing and backwards from what I'd expect. 

In addition, the colors seem a bit difficult to see (using Chrome on MacOS) and the mouseover effect is barely discernable.

Screenshot 2023-12-11 at 16 22 23

Looking great though! I imagine once Paseo launches this will get a lot of new visitors :)

### LINKED ISSUE #370: Paseo: configure the whitelist: instead of parity.io and w3f - users list

mordamax:
Closes #370 -> add paseo to UI, + whitelist paseo people separately from common pattern of parity and w3f
Closes #365 -> fix UX

### DIFF
--- a/client/src/lib/utils/networkData.ts
+++ b/client/src/lib/utils/networkData.ts
@@ -69,7 +69,7 @@ export const Trappist: NetworkData = {
 
 export const Networks: { network: NetworkData; url: string }[] = [
   { network: Rococo, url: (base as string) || "/" },
-  { network: Paseo, url: `${base as string}/paseo`},
+  { network: Paseo, url: `${base as string}/paseo` },
   { network: Westend, url: `${base as string}/westend` },
   { network: Trappist, url: `${base as string}/trappist` },
 ];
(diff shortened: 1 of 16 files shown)


client/src/lib/utils/networkData.ts
```diff
@@ -58,6 +69,7 @@ export const Trappist: NetworkData = {
 
 export const Networks: { network: NetworkData; url: string }[] = [
   { network: Rococo, url: (base as string) || "/" },
+  { network: Paseo, url: (base as string) || "/paseo" },
```
Bullrich, line 72: Shouldn't it be the following?

```suggestion
  { network: Paseo, url: `${base as string}/paseo`},
```

Bullrich (approved):
From the UI's side it seems good.

It seems that one of the tests is failing.

rzadp (approved):

PR #379, opened 2024-03-11
Using PAPI for e2e tests

mutantcornholio:
* Currently, e2e types are generated on postinstall and prebuild steps,
which requires metadata to be available in Docker image, even though
it's not technically used in the application code. I think separating
things would make things more complicated, but it's possible.
* Couldn't undestand the reason for having two different COPY instructions
in Dockerfile, but that doesn't live well with generating types in
postinstall, so merging those two
* .scale metadata files from Zombienet hosts saved in the codebase,
in order to be able to compile the code separately from running Zombienet


rzadp:
> Couldn't undestand the reason for having two different COPY instructions
in Dockerfile

This is an optimization technique, to cache a docker layer with `node_modules` which doesn't get rebuilt when source files get changed.

mutantcornholio:
> > Couldn't undestand the reason for having two different COPY instructions
> > in Dockerfile
> 
> This is an optimization technique, to cache a docker layer with `node_modules` which doesn't get rebuilt when source files get changed.

I see. After thinking more on this, decided to take this change back and go with another route.

e2e/README.md
rzadp, line 1: Can we add a note about how to obtain the `*.scale` files?

rzadp (approved):
🚀

Bullrich (approved):
LGTM

e2e/README.md
mutantcornholio: Good point, added

alvicsam (approved):

PR #392, opened 2024-04-12
PAPI for both application and test

mutantcornholio:


src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -1,223 +1,165 @@
...
+      parachain_id,
+      client,
+      nonce: this.#nonce++,
+    });
+
+    logger.debug(`Teleporting to ${address}: ${parachain_id}. Transaction ${tx}`);
+
+    // client.submit(tx) waits for the finalized value, which might be important for real money,
+    // but for the faucet drips, early respond is better UX
+    const submit$ = client.submitAndWatch(tx);
```
Bullrich, line 93: Thanks for the clarification here

src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -1,223 +1,165 @@
...
+      dripAmount,
+      address,
+      client,
+      nonce: this.#nonce++,
+    });
+    logger.debug(`Dripping to ${address}. Transaction ${tx}`);
+    // client.submit(tx) waits for the finalized value, which might be important for real money,
+    // but for the faucet drips, early respond is better UX
+
+    const submit$ = client.submitAndWatch(tx);
```
Bullrich, line 115:
```suggestion
    logger.debug(`Dripping to ${address}. Transaction ${tx}`);
    
    // client.submit(tx) waits for the finalized value, which might be important for real money,
    // but for the faucet drips, early respond is better UX
    const submit$ = client.submitAndWatch(tx);
```

Just a aesthetic change.

src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -1,223 +1,165 @@
...
         throw new Error(
-          `Can't send ${formattedAmount} ${networkData.currency}s, as balance is only ${formattedBalance} ${networkData.currency}s.`,
+          `Can't send ${formattedAmount} ${networkData.data.currency}s, as balance is only ${formattedBalance} ${networkData.data.currency}s.`,
         );
       }
 
       // start a counter and log a timeout error if we didn't get an answer in time
       dripTimeout = rpcTimeout("drip");
-      if (parachain_id != "") {
+      if (parachain_id !== null) {
```
Bullrich, line 144: Wouldn't `!!parachain_id` obtain the same result?

src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -1,223 +1,165 @@
...
-        result = { hash: hash.toHex() };
+        result = await this.transferTokens(amount, address);
       }
     } catch (e) {
-      result = { error: (e as Error).message || "An error occured when sending tokens" };
       logger.error("⭕ An error occured when sending tokens", e);
+      let message = "An error occured when sending tokens";
+      if (e instanceof Error) {
+        message = e.message;
+      }
```
Bullrich, line 154: I never thought of re assigning the message object. I'll copy this in my code.

src/papi/chains/index.ts
```diff
@@ -0,0 +1,70 @@
...
+export function getNetworkData(networkName: string): { data: NetworkData; api: NetworkApi } {
+  if (!Object.keys(networks).includes(networkName)) {
+    throw new Error(
+      `Unknown NETWORK in env: ${networkName}; valid networks are: [${Object.keys(networks).join(", ")}]`,
+    );
+  }
+  // networkName value is valdated one line before, safe to use as index
+  // eslint-disable-next-line security/detect-object-injection
+  return networks[networkName];
+}
```
Bullrich, line 70: Isn't the same result to do the following, or is there an advantage by looking for the key to exist?

```suggestion
export function getNetworkData(networkName: string): { data: NetworkData; api: NetworkApi } {
  const network = networks[networkName];
  if (!network) {
    throw new Error(
      `Unknown NETWORK in env: ${networkName}; valid networks are: [${Object.keys(networks).join(", ")}]`,
    );
  }
  return network;
}
```

Bullrich (commented):
Great PR, congratulations on finishing such a great rewrite 👏 

You kept it quite clear and most of it was a nice read.

I only have a couple of notes and a question.

---

Is it possible to recycle the code inside `src/papi/chains/*.ts`?

The code is quite similar and duplicated between 5 files so maybe it would make sense to abstract it for our forks.

client/.env
```diff
@@ -3,7 +3,7 @@ PUBLIC_CAPTCHA_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
 
 PUBLIC_FAUCET_URL=
 # uncomment to direct requests to local instance
-# PUBLIC_FAUCET_URL=http://localhost:5555/drip/web/
+ PUBLIC_FAUCET_URL=http://localhost:5555/drip/web/
```
rzadp, line 6:
```suggestion
PUBLIC_FAUCET_URL=http://localhost:5555/drip/web/
```

Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM docker.io/library/node:18.16.0-alpine
+FROM docker.io/library/node:20.11.1-alpine
 
 # uncomment to fix build on MacOS Apple Silicon chip
-# RUN apk add --no-cache python3 make g++
-RUN apk add git
+RUN apk add --no-cache python3 make g++
+# RUN apk add git
```
rzadp, line 5: Is `git` no longer needed here? I remember we had problems in CI without it.

src/bot/index.ts
```diff
@@ -126,7 +126,8 @@ bot.on(mSDK.RoomEvent.Timeline, (event: mSDK.MatrixEvent) => {
     logger.debug(`Processed receiver to address ${address} and parachain id ${parachain_id}`);
 
     try {
-      decodeAddress(address);
+      // TODO: check this
```
rzadp, line 129: What needs to be checked?

src/test/setupE2E.ts
```diff
@@ -204,31 +205,48 @@ async function setupAppContainer(params: {
...
+    SMF_CONFIG_DB_USERNAME: "postgres",
+    SMF_CONFIG_DB_PASSWORD: "postgres",
+    SMF_CONFIG_DB_DATABASE_NAME: "faucet",
+    // Public testing secret, will accept all tokens.
+    SMF_CONFIG_RECAPTCHA_SECRET: "6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe"
+  };
+
+  if (process.env.PAPI_DEBUG) {
+    env.PAPI_DEBUG = process.env.PAPI_DEBUG;
+    //console.log("PIDORASKA")
```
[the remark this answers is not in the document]
mutantcornholio: 😅

src/bot/index.ts
mutantcornholio: I've removed it locally, but forgot to commit it :\
I wanted to doublecheck that it actually works as an address validation (it does)

src/dripper/polkadot/PolkadotActions.ts
mutantcornholio: It's just reassigning a string variable. You sure you saw it the right way?

src/dripper/polkadot/PolkadotActions.ts
mutantcornholio: If I need to reduce `number | null` to `null`, I prefer to do it explicitly

src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -1,223 +1,165 @@
...
+    logger.info(`💸 teleporting tokens to ${address}, parachain ${parachain_id}`);
+
+    const addressBinary = Binary.fromBytes(encodeAccount(address));
+
+    const tx = await networkData.api.getTeleportTx({
+      dripAmount,
+      address: addressBinary,
+      parachain_id,
+      client,
+      nonce: this.#nonce++,
```
rzadp, line 86: Is it possible that:

1. We increase this nonce here.
2. We fail to send the transaction (for example a temporary network issue).
3. Upon a next request, we increase the nonce again.
4. The faucet is bricked, because it tries to send transactions with too high nonce?

mutantcornholio: Hmm, that's a good point actually. Let's update the nonce from the RPC periodically.

src/dripper/polkadot/PolkadotActions.ts
Bullrich: I was always doing something like:
```ts
if (e instanceof Error) {
    return e.message;
} else if (e instanceof CustomError) {
    return e.customMessage;
}
return "There was an error"
```

Now that I think about it's kind of the same code 😅 

Still, I hate error objects uncertainty in JS

src/dripper/polkadot/PolkadotActions.ts
mutantcornholio: There's a complication with that idea though. We can only synchronize nonce when there aren't any pending transactions, as the nonce gets updated when transactions are finalized.  
Although, a check for pending transactions inside of `PolkadotActions` shouldn't be hard to do.

Dockerfile
mutantcornholio: I believe it was needed when we had some git dependencies in `package.json`.
While we're at it, WDYT about adding `python3 make g++` by default, even it's only needed for local development?

rzadp: Sounds OK to me.

src/dripper/polkadot/PolkadotActions.ts
mutantcornholio: WDYT? https://github.com/paritytech/polkadot-testnet-faucet/pull/392/commits/7c2d0244f96139f9e11eac43f2caa3f7eeceff6e

rzadp: Looks good to me 👍

rzadp (approved):

src/papi/chains/index.ts
mutantcornholio: Your way looks cleaner, thanks!

AndWeHaveAPlan (approved):

PR #402, opened 2024-04-30
Moved GitLab checks to GitHub's action

### LINKED ISSUE #401: Migrate gitlab checks to github
Let's migrate the CI checks to work on GitHub actions so external contributors can see the results

image

Bullrich:
Resolves #401

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,8 @@
 name: Tests
 
 on:
+  push:
+    branches: ["main"]
   pull_request:
 
 permissions:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,16 +18,17 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 20
-      - name: Cache node modules
-        id: cache-npm
-        uses: actions/cache@v4.0.2
-        env:
-          cache-name: cache-node-modules
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "dir=$(yarn cache dir)" >> $GITHUB_OUTPUT
+
+      - uses: actions/cache@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,59 @@
+name: Tests
+
+on:
+  pull_request:
```
rzadp, line 4: Also push to `main` would be nice.

mutantcornholio (changes requested):
Can we have separate checks for lint and test please? 
If check fails, I'd like to know if it's only codestyle or the tests too.

Bullrich:
> Can we have separate checks for lint and test please? If check fails, I'd like to know if it's only codestyle or the tests too.

We have them:
image

`check-lint` checks the lint, and `check-test` checks the test.

If one fails the other one will still run

mordamax (approved):

mutantcornholio:
I just noticed something: upon submitting a review, the checks were restarted. We should probably select, which `pull_request` events we're listening for

Bullrich:
> I just noticed something: upon submitting a review, the checks were restarted. We should probably select, which `pull_request` events we're listening for

It was a coincidence, I just pushed.

By default, [it only uses `opened`, `synchronized` and `reopened`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) unless you specify.

mutantcornholio:
> > Can we have separate checks for lint and test please? If check fails, I'd like to know if it's only codestyle or the tests too.
> 
> We have them: image
> 
> `check-lint` checks the lint, and `check-test` checks the test.
> 
> If one fails the other one will still run

ah, sorry, my bad. I saw them in the single file, and thought that it'll result in the same check status

mutantcornholio (approved):

AndWeHaveAPlan (approved):

PR #405, opened 2024-05-14
Fixing "Invalid" error

mutantcornholio:
The source of the error was, that I treated the observables the wrong
way.
I expected `submitAndWatch()` to return a chain of events, to which I
could subscribe from a bunch of sources, while it actually led to side
effects being multiplied in the sources.
`shareReplay()` returns a multicast observable that actually behaves the
way I expected it to.

### DIFF
--- a/src/dripper/polkadot/PolkadotActions.ts
+++ b/src/dripper/polkadot/PolkadotActions.ts
@@ -1,6 +1,6 @@
 import { ss58Address } from "@polkadot-labs/hdkd-helpers";
 import { AccountId, Binary } from "polkadot-api";
-import { filter, firstValueFrom } from "rxjs";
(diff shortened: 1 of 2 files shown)


mordamax (approved):

src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -101,7 +101,7 @@ export class PolkadotActions {
 
     // client.submit(tx) waits for the finalized value, which might be important for real money,
     // but for the faucet drips, early respond is better UX
-    const submit$ = client.submitAndWatch(tx);
+    const submit$ = client.submitAndWatch(tx).pipe(shareReplay(1));
```
Bullrich, line 104: [TIL](https://rxjs.dev/api/operators/shareReplay)

PR #407, opened 2024-05-17
Fixing "Invalid" Error II

josepot:
I probably did a bad job at explanning how `shareReplay` works, my apologies.

The thing is that every time that with `shareReplay` every time that its internal refCount goes down to zero, then it unsubscribes from the source.

Therefore, there must be a subscription always open for the observable to be "shared". This is because observables are not eager. What was happening is that every time the old code was performing a `firstValueFrom` that creates a subscription and when the condition is met (and the promise resolves) then it unsubscribes from the observable. Since that was its only subscription, then the refcount goes down to 0 and the observable dies. So, the next `firstValueFrom` creates a brand new subscription.

I realize that people coming from PJS may find Observables difficult to deal with. So, we will probably add an optional callback to the promise based functions so that you can introspect what's going on under the hood without having to learn how to work with observables.

### DIFF
--- a/src/dripper/polkadot/PolkadotActions.ts
+++ b/src/dripper/polkadot/PolkadotActions.ts
@@ -112,19 +112,19 @@ export class PolkadotActions {
     // client.submit(tx) waits for the finalized value, which might be important for real money,
     // but for the faucet drips, early respond is better UX
     const submit$ = client.submitAndWatch(tx).pipe(shareReplay(1));
-
-    const hash = (await firstValueFrom(submit$.pipe(filter((value) => value.type === "broadcasted")))).txHash;
-
-    void firstValueFrom(submit$.pipe(filter((value) => value.type === "finalized")))
-      .catch((err) => {
-        logger.error(`Transaction ${hash} failed to finalize`, err);
-      })
-      .finally(() => {
+    let hash = "";
+    submit$.subscribe({
+      error: (err) => {
+        if (hash) logger.error(`Transaction ${hash} failed to finalize`, err);
         this.#pendingTransactions.delete(tx);
-      });
+      },
+      complete: () => {
+        this.#pendingTransactions.delete(tx);
+      },
+    });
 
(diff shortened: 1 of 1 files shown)


src/dripper/polkadot/PolkadotActions.ts
```diff
@@ -112,16 +112,21 @@ export class PolkadotActions {
     // client.submit(tx) waits for the finalized value, which might be important for real money,
     // but for the faucet drips, early respond is better UX
     const submit$ = client.submitAndWatch(tx).pipe(shareReplay(1));
+    let hash = "";
+    submit$.subscribe({
+      next: (e) => {
+        hash = e.txHash;
+      },
+      error: (err) => {
+        if (hash) logger.error(`Transaction ${hash} failed to finalize`, err);
```
mutantcornholio, line 121: Does it mean that `await firstValueFrom(...)` down below won't throw on errors?

josepot: it will throw, of course.

However, please notice that the previous code was not taking into account the fact that it could throw prior to being broadcasted, and in those cases the `this.#pendingTransactions.delete(tx);` was not being done.

josepot: basically, this subscription is the "equivalent" way of [doing this](https://github.com/paritytech/polkadot-testnet-faucet/pull/407/files#diff-801c8a922a2beb656a4bb54b0a2bb88c22032aa335dd983be04af6dc3a1b6282L118-L124), while also ensuring that the refcount doesn't go down to zero.

mutantcornholio: Then in case of `logger.error(`Transaction ${tx} failed to broadcast`, err);`, the error will get double-logged?

josepot: Oh, I see, you only want those logs to appear if an error happens after the promise of `sendTx` has already resolved. Gotcha! It's fixed now.

mutantcornholio (approved):

PR #412, opened 2024-05-22
Reuse matrix helpers from `@eng-automation/testing`

rzadp:


rzadp:
Running into some weird issues with `yarn install` with this.

mutantcornholio:
@rzadp well, one way around it is to remove `yarn.lock`

Bullrich:
We can also make eng-automation work both with es modules and with ES2020 for every case

rzadp:
> @rzadp well, one way around it is to remove `yarn.lock`

It helped locally, but not on CI 🤔

mutantcornholio (approved):

PR #426, opened 2024-06-21
Set paseo as default network

### LINKED ISSUE #424: Change default network to Paseo
Since the Rococo deprecation has been announced, users should see Paseo as the first network selected when they land on faucet.polkadot.io

Bullrich:
Resolves #424

I also added a new header text as requested by @lovelaced


lovelaced (approved):
Thanks, looks great

mutantcornholio (approved):

PR #428, opened 2024-08-14
Introduce reactions

rzadp:
Analogous to how the tip bot works, introduced reactions.

The user will get immediate feedback that the command has been received.

### DIFF
--- a/src/bot/index.ts
+++ b/src/bot/index.ts
@@ -48,6 +48,12 @@ const sendMessage = (roomId: string, msg: string, formattedMsg?: string) => {
   bot.sendEvent(roomId, null, "m.room.message", msgObject, "").catch((e) => logger.error(e));
 };
 
+const sendReaction = (roomId: string, eventId: string, emoji: string) => {
+  const msgObject: mSDK.IContent = { "m.relates_to": { event_id: eventId, key: emoji, rel_type: "m.annotation" } };
+
+  bot.sendEvent(roomId, null, "m.reaction", msgObject, "").catch((e) => logger.error(e));
+};
+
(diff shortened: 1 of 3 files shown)


mordamax (approved):
👀👍

src/bot/index.ts
```diff
@@ -114,8 +121,10 @@ bot.on(mSDK.RoomEvent.Timeline, (event: mSDK.MatrixEvent) => {
       logger.error("⭕ An error occurred when checking the balance", e);
     });
   } else if (action === "!drip") {
+    sendReaction(roomId, eventId, "👀");
     if (!arg0) {
-      logger.warn("Address not provided, skipping");
+      sendReaction(roomId, eventId, "😕");
+      sendMessage(roomId, `${sender} Please provide an address to be funded.`);
       return;
```
mutantcornholio, line 128: Sounds like there's no need to send 👀 reaction, if we're sending a 😕 right after.  
Why not move `sendReaction(roomId, eventId, "👀");` after synchronous checks?

rzadp:
Note: investigating the failing E2E tests. Not on issue with the PR, they fail on clean master as well.

src/bot/index.ts
rzadp: I see what you mean, also thought about it.

In my pedantic mind it's more consistent to always have the same pattern:

- `seen 👀` -> `👍 OK`
- `seen 👀` -> `😕 FAIL`

Changing this would introduce a `😕 FAIL` for a message that "has not been seen" in contrast with others.

Those were just my thoughts, don't know if it makes sense :D

mutantcornholio (approved):

PR #445, opened 2024-09-13
Full compat

mutantcornholio:
there's a theory that papi-related crashes could be fixed with enabling
whole compat package; let's see if that's the case


mutantcornholio:
I've deployed this change together with debug info (https://github.com/paritytech/polkadot-testnet-faucet/pull/446), and it seems to have helped westend at least

mordamax (approved):

PR #468, opened 2025-01-20
ESM setup and getting rid of Jest

### LINKED ISSUE #466: ESM setup
Even though, ESM transition is still a pain in the ass, more and more libraries support only ESM-projects. For example, `matrix-js-sdk`, that's been on dependabot's radar [for a long time now](https://github.com/paritytech/polkadot-testnet-faucet/pull/432).  

Let's start moving our projects, starting with faucet.

mutantcornholio:
Jest works terribly with ESM, needing its own weird setup for TS,
module name remapping and so on. Native test runner looks mature enough,
and even module mocks work fairly well.
Using native `assert` instead of jest's `expect` was a pain in the ass,
so opted for using modern assertion library `earl` instead.

* Importing JSONs from root works not as good, so moved config schema
inside the sources
* Yarn upgrade also solved some issue down the road
* Adding `.js` to all imports is a ESM requirement (even if it's
originally `.ts` files)
* mockHelpers.ts will probably be moved to `@eng-automation/testing`
* Replaced synchronous imports of `@eng-automation/testing` with
`import()`'s for now; will be properly addressed when with further
transitioning to ESM

Fixes #466


mordamax (approved):
🔥

PR #473, opened 2025-05-15
Westend AH migration

mutantcornholio:


client/tests/westend.test.ts
```diff
@@ -1,7 +1,6 @@
 import { FaucetTests } from "./faucet.js";
 
 const chains = [
-  { name: "Westend Relay", id: -1 },
   { name: "AssetHub", id: 1000 },
```
mordamax, line 4: in the file above you changed ah to -1? here too? or at least there should one be with -1?

mordamax (approved):

client/tests/westend.test.ts
mutantcornholio: That was one part of it, yes

PR #474, opened 2025-05-15
Update faucet for AHM on Westend - only drip on AH as transfer

seadanda:
Initially this was put up for reference to help the testing of the other solution, but at this point this should just be deployed ASAP.

### DIFF
--- a/client/src/lib/utils/networkData.ts
+++ b/client/src/lib/utils/networkData.ts
@@ -42,11 +42,7 @@ export const Westend: NetworkData = {
   networkName: "Westend",
   currency: "WND",
   chains: [
-    { name: "Westend Relay", id: -1 },
-    { name: "AssetHub", id: 1000 },
-    { name: "Collectives", id: 1001 },
-    { name: "BridgeHub", id: 1002 },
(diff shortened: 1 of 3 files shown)


client/src/lib/utils/networkData.ts
```diff
@@ -42,10 +42,7 @@ export const Westend: NetworkData = {
   networkName: "Westend",
   currency: "WND",
   chains: [
-    { name: "Westend", id: -1 },
-    { name: "Collectives", id: 1001 },
-    { name: "BridgeHub", id: 1002 },
-    { name: "People", id: 1004 },
+    { name: "Westend", id: 1000 },
```
seadanda, line 45: This is here for compatibility. On the UI the chains dropdown should be totally removed when Westend is selected (or just in general)

mutantcornholio:
As suggested teleport logic doesn't work, and there's a general consensus about removing teleports for Westend (which implies a substantially bigger change), I'm closing this PR, and will finish the task in https://github.com/paritytech/polkadot-testnet-faucet/issues/472

PR #482, opened 2025-07-23
Decrease paseo drips

mutantcornholio:
Paseo faucet drains too quickly, and 5k PAS is a lot of tokens, which
aren't needed for most operations.


mordamax (approved):
Lets confirm with pas team though

PR #484, opened 2025-08-21
Paseo faucet to run on Asset Hub instead of Relay

karolk91:
Following PR consists of:
- Switch faucet to use Paseo Asset Hub chain as source / pot instead of Relay
- Support teleports to relay network from parachain
- - a special value for `parachain` variable == `0` means "parent" / "relay" from the point of view of parachain
- - a special value for `parachain` variable == `-1` means `this network`, so in such case
- Introduce `id` attribute to `NetworkData` to be able to recognize transfer funds to self (i.e. when `parachain` == `id` then use simple balance transfer , instead of teleport to `self` which is invalid op)

### Todo:
- [ ] fix Matrix bot to automatically set `parachain=0` parameter in case of empty `:ParachainId` in case of `!drip` command
- [ ] Introduce tests for parachain == 0
- [ ] Check E2E tests


karolk91:
> Will teleporting from AH to PassetHub work? I assume not, judging by the code. We definitely need it, and the suggestion from @seadanda was to use Paraspell SDK for teleporting instead of PAPI.

Hi @mutantcornholio, teleport to Passet Hub works just fine, I have tested it - its no different than teleport to any other system chain. Why would you think its not going to work ?

As per @seadanda suggestion to use Paraspell, I totally agree. But as far as I understand we need faucet to work post-AHM on Paseo asap (as the migration is very close). Incorporating Paraspell will require siginificatly more changes, so we can think of this solution herer as intermediate one with the end goal to transition to Paraspell

mutantcornholio:
> Hi @mutantcornholio, teleport to Passet Hub works just fine, I have tested it - its no different than teleport to any other system chain. Why would you think its not going to work ?

After migrating Westend to AH, teleporting from AH to other parachains broke (or rather, didn't work, as we were teleporting from relay chain before), and @seadanda suggested a wildly different and much more complicated XCM call for that case, which also didn't work, so we've decided to disable parachain drips altogether.  

So you're saying that dripping to PassetHub works with these changes, even if faucet account is on AssetHub?

karolk91:
> So you're saying that dripping to PassetHub works with these changes, even if faucet account is on AssetHub?

Yes, I have started local network using chopstics (Paseo Relay, Paseo AH, Paseo Passet Hub, Paseo People) with pre-funded Alice account. Also started local faucet (and used Alice account as pot) and I can sucessfuly teleport from Paseo AH to any of these (also to Paseo AH itself)

Surely, my local setup is not the "post-AHM" environment, but I'm not aware of anything that would break this behavior

karolk91:
Let me actually check this implementation against Westend. I dont see why it wouldn't work

seadanda:
> fix Matrix bot to automatically set parachain=0 parameter in case of empty :ParachainId in case of !drip command (this is to be compatible with previous behavior, when user wanted some funds and didn't pass Parachain arg, he expected to receive funds on relay network

The default should switch to dripping on AH, since that's where all the functionality is moving. We shouldn't support dripping to the relay chain post-AHM

I've removed paraspell from the issue btw, since it's already working with minimal changes

PR #513, opened 2026-05-12
skip cap check if not main AH

mordamax:
Fixes

image

### DIFF
--- a/src/dripper/DripRequestHandler.ts
+++ b/src/dripper/DripRequestHandler.ts
@@ -1,8 +1,12 @@
+import { config } from "#src/config";
 import { isDripSuccessResponse } from "#src/guards";
 import { logger } from "#src/logger";
 import { counters } from "#src/metrics";
+import { getNetworkData } from "#src/papi/index";
 import { DripRequestType, DripResponse, TxStatusCallback } from "#src/types";
 import { isAccountPrivileged } from "#src/utils";
+
+const sourceChainId = getNetworkData(config.Get("NETWORK")).data.id;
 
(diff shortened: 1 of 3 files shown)


src/dripper/DripRequestHandler.ts
```diff
@@ -53,7 +53,10 @@ export class DripRequestHandler {
 
     const isAllowed = !(await hasDrippedToday(external ? { addr } : { username: opts.sender, addr }));
     const isPrivileged = !external && isAccountPrivileged(opts.sender);
-    const isAccountOverBalanceCap = await this.actions.isAccountOverBalanceCap(addr);
+    // Cap check uses the faucet's source-chain balance. Skip it for teleports — the destination
+    // chain's balance is not reachable from the faucet's single RPC connection.
+    const isTeleport = validatedParachainId !== null;
```
UtkarshBhardwaj007: Is this too permissive? Since `validateParachainId` accepts 1000 a caller can submit `parachain_id=1000` to skip the cap check while still receiving a regular in-chain transfer to a rich Asset Hub account. The UI's custom-chain input makes this possible I guess.

UtkarshBhardwaj007: Maybe add a test for this too.

UtkarshBhardwaj007 (approved):
Approving to unblock, please address comment before merging.

src/dripper/DripRequestHandler.ts
mordamax: great catch, fixed! thanks!

PR #520, opened 2026-06-09
Add "Make another request" button to web drip success screen

shawntabrizi:
## Problem

After a successful drip from the web UI, the success screen showed only the confirmation message (and a transaction link, where available) with **no way back to the form**. Users had to reload the page to make another request.

## Fix

Mirror the existing `Error.svelte` "Go back" pattern: add a **"Make another request"** button to `Success.svelte` that resets the `operation` store to `undefined`, which returns the user to the drip form.

- When a transaction-explorer link is present, the reset renders as a **secondary** button so it doesn't compete with the primary "See transaction details" link.
- When there's no explorer link, it uses the primary `submit-btn` style.
- Added a shared `.secondary-btn` style in `app.css` next to `.submit-btn` (button styles live in global CSS here).

## Testing

- Added a Playwright test verifying the button appears after a successful drip and clicking it returns to a usable form.
- Existing "display link to transaction" test still passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/client/src/app.css
+++ b/client/src/app.css
@@ -177,6 +177,25 @@ div.form-field:focus-within {
   border: 1px solid #44403c;
 }
 
+/* ── Secondary button ── */
+.secondary-btn {
+  @apply btn w-full rounded-xl;
+  height: 2.75rem;
+  margin-top: 0.75rem;
+  background-color: transparent;
+  color: #a8a29e;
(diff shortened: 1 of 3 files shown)


mordamax (approved):
nice!

